PR #13, opened 2020-01-24
Give R access to Whitney Book font

nmpeterson:
I believe this fixes the font issue -- it worked for me -- but someone else needs to test on their PC (assuming you have a font called "Whitney Book" installed) to verify.

Sample output:
image

### DIFF
--- a/R/theme_cmap.R
+++ b/R/theme_cmap.R
@@ -1,7 +1,8 @@
 #' Add CMAP theme to ggplot chart
 
-style_cmap <- function() {
-  font <- "Whitney"
+theme_cmap <- function() {
+  windowsFonts(`Whitney Book` = "TT Whitney Book")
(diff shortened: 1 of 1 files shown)


tallishmatt:
This works, but I think per Comms the CMAP font family is our official font now, rather than Whitney. There is some difference:
image

We may also want to consider incorporating a backup font if our preferred font is unavailable.

nmpeterson:
Ok, then I'm guessing CMAP Sans is just a renamed Whitney Light. Either way, we should double-check with Comms as to what typeface and weight(s) they want us using in charts -- titles may need to be bold/semibold, for example.

tallishmatt:
Yup! There is still supposedly an updated CMAP visual guidelines document coming down the pipe at some point, but if that is still a ways off we might just need to sit down with Nancy.

PR #19, opened 2020-01-29
Palettes

gritzenthaler:
A few discrete palettes and race palettes. Will add more discrete and continuous soon.


gritzenthaler:
Going to resubmit pull request with continuous palettes included

PR #23, opened 2020-02-03
Move font specification into global variables

nmpeterson:
Also improved documentation


nmpeterson:
Hold off on merging this for now. I think I've found a solution allowing the specification of a backup font that I will try to add to this.

nmpeterson:
Never mind, the solution I thought I found turned out to be unreliable. :(

Please review the code and merge.

tallishmatt:
This appears to contain both font-fix AND label_last_point related changes. Is this meant to be both? The font-fix aspects look great, but I do have some questions about the label_last_point feature.

nmpeterson:
What questions?

The bulk of the label-last-point functionality got merged into master accidentally a week or two ago because GitHub's desktop app forced me into doing that when I was merging master into label-last-point. The only code unrelated to fixing fonts in this PR is some extra documentation.

tallishmatt:
Ah, forgot about that.

Questions about label_last_point mostly pertain to implicitly automating behavior that the user is likely to need pretty much every time this function is added. Of course, this means more work now, and it may be that some of these things are not worth the effort. But here's what comes to mind:

1. Could it be programmed in such a way that defaults to taking the Y axis value as the label, rather than requiring the user to set the label attribute in top-line aes?  Considering that the following two code blocks yield the same output, can `geom_text_lastonly()` be set up to inherit the master aes y value as it's label value?

Label in topline aes:
```ggplot(grp_over_time, aes(x = year, y = realgrp, color = cluster, label = realgrp)) +
    scale_y_continuous("Gross regional product (indexed to 2007)", labels=scales::percent) +
    scale_x_continuous("Year") +
    geom_line() + geom_text_lastonly()
```
label in secondary aes:
```ggplot(grp_over_time, aes(x = year, y = realgrp, color = cluster)) +
    scale_y_continuous("Gross regional product (indexed to 2007)", labels=scales::percent) +
    scale_x_continuous("Year") +
    geom_line() + geom_text_lastonly(aes(label = realgrp))
```

2. Similar question re formatting: could the formatting applied to the y axis apply to the label by default? 

3. in my test plot, which involves a legend, the legend gets modified by the addition of the label. See the image below. 

4. I forget where we left this off in our discussion last week - considering so much of the functionality is the same, is it worth incorporating adding a dot on the last point into the same function?

5. is it within reason to automate the `expand_scale()` so it occurs within this function?

image

tallishmatt:
updated the above comment a moment ago to correct display error in code blocks.

nmpeterson:
I created a new issue (#24) with your suggestions. They don't relate to fonts and should be kept separate from this PR.

PR #25, opened 2020-02-11
Palettes

gritzenthaler:


nmpeterson:
@gritzenthaler Please double-check all of your gradients. I noticed several that had the wrong hex codes included. Also, every multi-hue gradient should have 5 colors, and every single-hue color should have 9 (and 17 for the grays). This is to avoid interpolation as much as possible, as the 3-color specification you've used causes intermediate colors to deviate pretty significantly from the actual CMAP palette.

nmpeterson:
Discrete palettes will need to be updated with Nancy's input, but otherwise I think this is looking good. Note that I did rename a few functions/parameters for brevity and consistency with `ggplot2` terminology.

PR #31, opened 2020-02-28
Use Calibri as backup font in case Whitney is not installed

### LINKED ISSUE #30: Set backup font in case Whitney is not installed
Use Calibri if possible.

nmpeterson:
I think this resolves #30. If the user does not have Whitney Medium, Whitney Book and Whitney Semibold all installed, it will use Calibri, Calibri Light, and Calibri (bold variant) instead, respectively.

Important note for @MMenninger @sarahcmap: since Calibri does not have a standalone bold font, but rather a bold variant of the regular-weight font, all `ggplot` text elements *must* now have both the `family` and `face` arguments specified.

For example:

    plot.title = ggplot2::element_text(family=cmapplot_globals$font_title,
                                       face=cmapplot_globals$font_title_face,
                                       size=28, color="#222222")

Once this pull request is approved, I highly recommend merging the changes into your theme development branches for testing.


matthewstern:
@nmpeterson this looks awesome. I do have a few questions though.
1. you adjusted the fontface in geom_text_lastonly in this PR. I though you had previously suggested that geom_text should automatically inherit default fonts. Should I adjust the recessions_text to include this specification?
2. in `devtools::check()` we get a warning saying `GeomTextLast` needs documentation. That appears to be because the function is exported. I would think it doesn't need to be exported, as it is not used by the user, only by `geom_text_lastonly`, which is in the same R file. Does that sound right?
3. How much documentation do we want to do, generally, to code? Looking at your now more complex font specification, it's complex enough that I can imagine future developers of this package benefiting from in-line commenting or readme-type documentation, even though that code isn't really user-facing. What do you think?

nmpeterson:
@matthewstern:

1. Don't change anything. This change will be reverted in a forthcoming PR.
2. This has also been addressed in the forthcoming PR.
3. More documentation is always better. I'll try to update this soon.

Issue #24, opened 2020-02-10
geom_text_lastonly() enhancements?


nmpeterson:
Questions about label_last_point mostly pertain to implicitly automating behavior that the user is likely to need pretty much every time this function is added. Of course, this means more work now, and it may be that some of these things are not worth the effort. But here's what comes to mind:

- [x] 1. Could it be programmed in such a way that defaults to taking the Y axis value as the label, rather than requiring the user to set the label attribute in top-line aes?  Considering that the following two code blocks yield the same output, can `geom_text_lastonly()` be set up to inherit the master aes y value as it's label value?

- [ ] ~~2. Similar question re formatting: could the formatting applied to the y axis apply to the label by default?~~

- [x] 3. in my test plot, which involves a legend, the legend gets modified by the addition of the label. See the image below. 

- [x] 4. I forget where we left this off in our discussion last week - considering so much of the functionality is the same, is it worth incorporating adding a dot on the last point into the same function?

- [ ] ~~5. is it within reason to automate the `expand_scale()` so it occurs within this function?~~

- [x] 6. Add optional `point_aes` and `text_aes` parameters to allow separate control of label/point geom parameters.

![image](https://user-images.githubusercontent.com/54633946/74174597-d0307d00-4bf9-11ea-84f9-be20950768f1.png)

_Originally posted by @tallishmatt in https://github.com/CMAP-REPOS/cmapplot/pull/23#issuecomment-584242491_

nmpeterson:
Regarding suggestion 3, if `show.legend = FALSE`, then the text labels will not be added to the legend. In pull request #31 I changed this to the default behavior, but it could be overridden by specifying `show.legend = TRUE` if desired.

nmpeterson:
As far as I can tell, suggestion 2 is impossible because the axis label formatting is specific to `scale_y_continuous` and is therefore not inheritable by `geom_text_lastonly`.

For similar reasons, suggestion 5 also seems impossible.

I am working on some code to address suggestions 1 & 4.

nmpeterson:
Consider modifying suggestion 6 with approach mentioned by @matthewstern in #41 (implemented for geom_recessions.R), with entirely separate `mapping`/`aes` parameters available for points.

PR #33, opened 2020-03-04
geom_text_lastonly() enhancements

nmpeterson:
Addresses the remaining enhancements (suggestions 1 & 4) in #24.


matthewstern:
@nmpeterson how do you feel about me changing the name of this function to `highlight_last_point`?

nmpeterson:
I'm not a fan of that idea. I feel it's important to keep the naming consistent with other `ggplot2` geoms, which indicates to users that it *is* a `ggplot2` geom and it *can* accept the common parameters like `color`, `size`, etc., as well as inherit parameters from a base `ggplot()` object. The current name also indicates its direct relationship to `geom_text`.

matthewstern:
Fair enough. My concern is that the function can now also add a geom_point layer as well, so I felt like the reference to geom_text is potentially misleading. You're not worried about that?

nmpeterson:
Not really, since that's not the default behavior and I think the `add_point` parameter name is pretty self-explanatory. The only thing of concern to me is that explicitly setting a `size` value affects both, but adding an optional `point_size` parameter would resolve that.

matthewstern:
OK, sounds good.

Issue #14, opened 2020-01-27
Incorporate sample ggplot code into README.Rmd; eliminate sample_graphics.R


matthewstern:
@nmpeterson per your request to document the sample datasets I added, I attempted to follow the instructions you sent [here](https://r-pkgs.org/data.html#documenting-data), which led me to chapter 8 of the same book and to the [basics of Roxygen](https://cran.r-project.org/web/packages/roxygen2/vignettes/roxygen2.html). However, I am lost and could use your help.

I am working in the 'adtl-sample-data' branch. I have code in the file data.R that should generate a man file, but when I attempt to build the man files by running `devtools::document()` from the console, I get an error that has to do with sample_graphics.R:  
`Error in theme_cmap() (from sample_graphics.R#8) : could not find function "theme_cmap"`

Would appreciate your troubleshooting help so that we can figure out how to use roxygen for all documentation, as that seems to be a common method/best practice for documentation. 

Thank you!

nmpeterson:
Try commenting out the entirety of sample_graphics.R while you're working on your documentation.

As a longer-term modification, I think that code makes more sense as part of README.Rmd than as a script in the R folder. This will have the effect of always prominently displaying the latest sample plots on the main page of the GitHub site.

tallishmatt:
OK, this is commented out in pull request #15 but I'm leaving this issue open as a reminder to eventually build the sample ggplot code into README.Rmd.

matthewstern:
OK this is done as of PR #35 .

PR #35, opened 2020-03-05
Readme edits

matthewstern:
Readme contains examples from sample_graphics.R; this file is now deleted.

Additionally, readme contains additional details about what example datasets are included and about how to call palettes. 

This addresses #14 .


nmpeterson:
Any objections if I set `eval=TRUE` for the sample code so that the different chart types will actually be displayed on the readme, as they are for the palettes?

matthewstern:
The only reason I didn't set it that way initially was that they felt really large. Also I figured the truly illustrative stuff would end up in the cookbook anyway.  But I definitely have no objection to the change for now.

nmpeterson:
Agreed that eventually the sample plots can be migrated from readme into cookbook (which should of course be linked from the readme).

PR #36, opened 2020-03-06
Added diverging gradients

nmpeterson:
Also modified lightest/darkest values from existing gradients, as they were too light/dark. Note that gradient names were also changed: sequential gradients have a "seq_" prefix, while diverging gradients have a "div_" prefix.


matthewstern:
This looks really good to me. Due to good timing from our Wednesday R group, I rewrote the readme file to iterate over the lists of palettes and gradients using the walk function. More elegant, good practice, plus changes to gradients will automatically be incorporated into the readme. Note that is required changing the prefix on multi-hue sequential gradients from seq* to multi*.

The readme currently does not capture the race palette in any way. We could solve this now or later. It looks like the race colors were written into the `cmap_fill_race` and `cmap_color_race` functions, so I think they are not globally available. Do y'all think it would be preferable to shift the color specification over to the palettes_discrete.R file, so that the colors are all set together and that gradient becomes globally available?

README.Rmd
```diff
@@ -136,43 +130,30 @@ percentile_wages %>%
...
-viz_gradient(cmap_gradients$seq_yellow_orange_red)
-viz_gradient(cmap_gradients$seq_green_teal_blue)
-viz_gradient(cmap_gradients$seq_yellow_orange)
-viz_gradient(cmap_gradients$seq_orange_red)
-viz_gradient(cmap_gradients$seq_red_purple)
-viz_gradient(cmap_gradients$seq_yellow_green)
-viz_gradient(cmap_gradients$seq_green_teal)
-viz_gradient(cmap_gradients$seq_teal_blue)
+cmap_gradients %>% 
+  keep(str_detect(names(.), "^multi")) %>%
```
nmpeterson: Can we revert the names and distinguish between single/multi hue gradients with a more nuanced regex? Specifically, multi-hues have more than 1 underscore, whereas single-hues only have one. I think the regex `^seq_[^_]+$` would work for single-hues and `^seq_[^_]+_.+$` would work for multi-hues.

gritzenthaler:
I can redefine the race palette outside of the function right now and add it to the readme. 

I only briefly looked into centering divergent palettes on zero, but I'm wondering if it's more of a user-end solution. What if we had an example in the documentation that showed how to do that?

nmpeterson:
@gritzenthaler 

> I can redefine the race palette outside of the function right now and add it to the readme.
> 
> I only briefly looked into centering divergent palettes on zero, but I'm wondering if it's more of a user-end solution. What if we had an example in the documentation that showed how to do that?

The ventering should always be customizable, even if we were to set a default of 0, because there will be times when someone wants the center to represent a median value, etc.

README.Rmd
matthewstern: Sure...I did consider a more complicated regex, but counting underscores seemed like a less logical and easier to accidentally break solution. This isn't really important to me either way, but why not have a distinguishing prefix?

nmpeterson: When I named them I felt it was important to distinguish between sequential and diverging (i.e. the gradients' use cases) rather than number of hues, because the diverging gradients are all multi-hue as well, so this could add confusion.

If you're worried about something breaking (which shouldn't happen as this list of gradients will be pretty static going forward, and has a consistent naming scheme regarding underscores), I would propose instead collapsing the single- and multi-hue sequential gradients into a single group on the readme that could all be identified with the regex `^seq`.

PR #43, opened 2020-04-07
Small tweaks to format of theme_cmap file

matthewstern:
created font warning if Whitney not available
shifted to do.call format so that list of theme attributes can be constructed dynamically
set generic font size


MMenninger:
will try to look at this this afternoon.

matthewstern:
Nevermind. Not yet.

PR #45, opened 2020-04-29
structural changes to theme_cmap file

matthewstern:
This PR does not solve many specific issues in theme formatting, but does suggest a new structure for how the theme file might work. A list of ggplot elements is introduced. The first list element is the primary theme, but additional list elements are added as needed based on function arguments.

If we're happy with this concept, after merging contributors should be able to add/modify various components of the file to bring the theme in alignment with our design goals expressed in #8.


nmpeterson:
From a brief look at the code, I think this is a very sensible improvement. I will try to look more closely later this week.

One small suggestion in the meantime: move the test plot code from the end of the script into Roxygen documentation for `theme_cmap()`. This is a holdover from before we learned about Roxygen.

matthewstern:
Thanks @nmpeterson. 

I merged master into this branch to update it and made some documentation tweaks, including shifting the example from a comment into the roxygen documentation. 

The original example relies on the pipe to run, so it breaks without `\dontrun{}` tags. I created another example that should work without `dontrun` but is throwing an error in `devtools::check()` that seems to have something to do with the `font_reg` variable in `cmapplot.R`. Not sure what to do with that.

PR #46, opened 2020-05-01
Add text_aes & point_aes params to geom_text_lastonly

nmpeterson:
Resolves suggestion 6 in #24

### DIFF
--- a/R/cmapplot.R
+++ b/R/cmapplot.R
@@ -42,7 +42,7 @@ if (.Platform$OS.type == "windows") {
       font_sbold = "Whitney Semibold"
     )
   } else {
-    warning("Whitney is not installed on this PC, so CMAP theme will default to Calibri")
+    message("Whitney is not installed on this PC, so CMAP theme will default to Calibri")
     grDevices::windowsFonts(
       sans = "Calibri",  # Override the default font (Arial)
       font_reg = "Calibri",
@@ -70,7 +70,7 @@ if (.Platform$OS.type == "windows") {
 } else {
 
   # Assume no Whitney or Calibri on non-Windows (i.e. non-work) computers
-  warning("CMAP theme will default to Arial on non-Windows platforms")
+  message("CMAP theme will default to Arial on non-Windows platforms")
   cmapplot_globals$use_whitney = FALSE
 
   grDevices::X11Fonts(
--- a/R/cmapplot.R
+++ b/R/cmapplot.R
@@ -42,7 +42,7 @@ if (.Platform$OS.type == "windows") {
       font_sbold = "Whitney Semibold"
     )
   } else {
-    message("Whitney is not installed on this PC, so CMAP theme will default to Calibri")
+    message("WARNING: Whitney is not installed on this PC, so CMAP theme will default to Calibri")
     grDevices::windowsFonts(
       sans = "Calibri",  # Override the default font (Arial)
       font_reg = "Calibri",
@@ -70,7 +70,7 @@ if (.Platform$OS.type == "windows") {
 } else {
 
   # Assume no Whitney or Calibri on non-Windows (i.e. non-work) computers
-  message("CMAP theme will default to Arial on non-Windows platforms")
+  message("WARNING: CMAP theme will default to Arial on non-Windows platforms")
   cmapplot_globals$use_whitney = FALSE
 
   grDevices::X11Fonts(
@@ -105,5 +105,3 @@ check_cmap_fonts <- function() {
   graphics::text(1, 1, "font_label", cex=4)
 }
 #check_cmap_fonts()
-
-ggplot2::update_geom_defaults("text", list(family=cmapplot_globals$font_main, face=cmapplot_globals$font_main_face))
--- a/R/geom_recessions.R
+++ b/R/geom_recessions.R
@@ -80,7 +80,7 @@
 #'   scale_x_date("Year") +
 #'   theme_minimal()
 #'
-#' # MODIFIED AESTHETICS
+#' # MODIFIED AESTHETICS:
 #' ggplot(grp_over_time, aes(x = year, y = realgrp)) +
 #'   geom_recessions(show.legend = TRUE, fill = "blue", text = FALSE,
 #'                   rect_aes = list(alpha = 1, color = "red")) +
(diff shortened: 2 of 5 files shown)


R/geom_recessions.R
```diff
@@ -1,352 +1,345 @@
-#'Add recessions to time series graphs
```
matthewstern, line 1: Geom_recessions and geom_text_lastonly files are showing up as completely rewritten in the compare for some reason. Can you briefly note any major changes? I can see the diff correctly in the .Rd help files, so it's really only a question pertaining to the code itself.

R/geom_recessions.R
```diff
@@ -1,352 +1,345 @@
...
+#' ggplot(grp_over_time, aes(x = year, y = realgrp, color = cluster)) +
+#'   geom_recessions() +
+#'   geom_line() +
+#'   scale_x_continuous("Year") +
+#'   theme_minimal()
+#'
+#' # DATE X AXIS:
+#' ggplot(data = grp_goods,
+#' mapping = aes(x = year2, y = realgrp, color = cluster)) +
+#'   geom_recessions(xformat = "date", ) +
```
matthewstern, line 78: unnecessary comma in this line. Kinda surprised that doesn't break things.

R/geom_recessions.R
```diff
@@ -1,352 +1,345 @@
...
+#' ggplot(grp_over_time, aes(x = year, y = realgrp)) +
+#'   geom_recessions(show.legend = TRUE, fill = "blue", text = FALSE,
+#'                   rect_aes = list(alpha = 1, color = "red")) +
+#'   geom_line(aes(color = cluster)) +
+#'   scale_x_continuous("Year") +
+#'   theme_minimal()
+#'
+#'
+#' # BELOW EXAMPLES SHOW MORE THAN 1 RECESSION
+#' df <- data.frame(date_dec=1800:2020, value=rnorm(442), var=c(rep("A", 221), rep("B", 221)))
```
matthewstern, line 93: Let's simplify this to a more even number of years! The examples use a 4 year spread. If we change this to 1950 to 1999, we can remove the filter lines below.

R/geom_recessions.R
```diff
@@ -1,352 +1,345 @@
...
+geom_recessions <- function(xformat = "numeric",
+                            text = TRUE,
+                            label = " Recession",
+                            ymin = -Inf,
+                            ymax = Inf,
+                            fill = "#002d49",
+                            text_nudge_x = 0.2,
+                            text_nudge_y = 0,
+                            show.legend = FALSE,
+                            ...,
```
matthewstern, line 129: `...` should go last I think.

R/geom_text_lastonly.R
```diff
@@ -1,310 +1,320 @@
...
+#'   geom_line() +
+#'   labs(title="Random lines") +
+#'   scale_y_continuous("Percentage of absolutely nothing", labels=scales::percent) +
+#'   scale_x_continuous("Year", expand=expansion(mult=c(0.05, 0.10))) +
+#'   geom_text_lastonly(add_points=TRUE, text_aes=list(fontface="bold"), point_aes=list(size=2.5))
+#'
+#' @export
+geom_text_lastonly <- function(mapping = NULL, data = NULL,
+                      stat = "identity", position = NULL,
+                      ...,
```
matthewstern, line 63: `...` should go last, I think?

R/geom_recessions.R
matthewstern: I fixed.

R/geom_recessions.R
matthewstern: I fixed.

R/geom_recessions.R
matthewstern: I moved to last position. @nmpeterson OK?

R/geom_text_lastonly.R
matthewstern: I moved to last position. @nmpeterson OK?

R/cmapplot.R
```diff
@@ -70,7 +70,7 @@ if (.Platform$OS.type == "windows") {
 } else {
 
   # Assume no Whitney or Calibri on non-Windows (i.e. non-work) computers
-  warning("CMAP theme will default to Arial on non-Windows platforms")
+  message("CMAP theme will default to Arial on non-Windows platforms")
   cmapplot_globals$use_whitney = FALSE
 
   grDevices::X11Fonts(
```
matthewstern, line 76: @nmpeterson this is unrelated to this PR, but this is my first time taking a close look at cmapplot.R in a while. I am curious if this line (76) or line 109 changes the default behavior of ggplot or other aspects of a user's interaction with their local R installation? I feel that ideally the installation or loading of this package should not change ggplot defaults, if we can avoid it. Thoughts?

R/geom_recessions.R
nmpeterson: Ugh, that's annoying. I made the changes on macOS, so the line endings must have been changed from the Windows to Mac/Unix line endings for all the files I edited. Shouldn't impact functionality at all, but does make it harder to review changes. For `geom_recessions`, all I did was simplify the example code. For `geom_text_lastonly`, I added `text_aes` and `point_aes` parameters, behaving exactly as the `text_aes`/`rect_aes` you added to `geom_recessions`. I also tweaked the example code to demonstrate this new functionality.

R/cmapplot.R
```diff
@@ -42,7 +42,7 @@ if (.Platform$OS.type == "windows") {
       font_sbold = "Whitney Semibold"
     )
   } else {
-    warning("Whitney is not installed on this PC, so CMAP theme will default to Calibri")
+    message("Whitney is not installed on this PC, so CMAP theme will default to Calibri")
```
matthewstern, line 45: I changed these from `warning` to `message`. 
Motivation: I don't like how warning messages take two lines.
Justification: we're not [fundamentally worried about how the package will function](https://www.r-bloggers.com/a-warning-about-warning/) if this error is produced, nor do we want the package to error in an environment with a [heightened warning setting](https://www.rdocumentation.org/packages/base/versions/3.6.2/topics/warning).

R/geom_text_lastonly.R
```diff
@@ -1,310 +1,322 @@
...
+                      nudge_x = 0.25,
+                      nudge_y = 0,
+                      check_overlap = FALSE,
+                      na.rm = FALSE,
+                      show.legend = FALSE,
+                      inherit.aes = TRUE,
+                      add_points = FALSE,
+                      text_aes = NULL,
+                      point_aes = NULL,
+                      ...)
```
matthewstern, line 75: I moved `...` to last position in this file and in `geom_recessions`. I can't imagine this is too big a deal but flagging for Noel in case you have concerns.

R/cmapplot.R
nmpeterson: I think we can discuss it. The reason I didn't have any qualms with it is that I figured people would only be loading this package when they're working on a plot for CMAP, and therefore need those changes to be reflected one way or another. But yeah, probably not "best practices"

matthewstern:
Sorry for the barrage of comments. I'm good with this if you're good.

R/cmapplot.R
nmpeterson: Can we change it to `message("WARNING: Whitney is not ...` in that case? I do want users to feel like they are being warned rather than just brushing these massages off.

R/cmapplot.R
matthewstern: See line 45 of `theme_cmap.R`. All text in the geom gets changed to the correct font face whenever `theme_cmap()` is called. I would suggest we can therefore delete line 109.

I'm not sure I understand `grDevices::X11Fonts` though.

R/cmapplot.R
matthewstern: Works for me.

R/geom_text_lastonly.R
nmpeterson: The positioning for all of the `...` parameters found in `geom_text_lastonly` was identical to the positioning in the original `ggplot` code that I copied. I personally would be inclined to leave them where they were before, but I won't put up a fight over it.

matthewstern: OK, I am reverting.

R/cmapplot.R
nmpeterson: Yes, that makes sense; I'm fine with deleting the final line of `cmapplot.R`.

`grDevices::X11Fonts` has to be used on Mac/Linux instead of `grDevices::windowsFont` to manage the fonts used by the package.

PR #48, opened 2020-08-31
Transit ridership sample data

dlcomeaux:
Upload sample data set and documentation for transit ridership over time. This includes the ridership data (transit_ridership.RData), changes to the data.R file to include documentation, and the outputted .Rd file in the "man" folder (created using devtools).


dlcomeaux:
Closing to create new request with multiple sets.

dlcomeaux:
Upload sample data set and documentation for transit ridership over time and vehicle ownership in the seven county region. 

This includes two .RData files, two additional sections in the data.R file, and two new files in the "man" folder that provide oxygen-based documentation.

R/geom_text_lastonly.R
```diff
@@ -52,7 +52,7 @@
 #'   geom_text_lastonly()
 #'
 #' # With points, label formatting and x-axis expansion
-#' ggplot(df, aes(x=year, y=value, color=var, label=sprintf("%.1f%%", 100*var))) +
+#' ggplot(df, aes(x=year, y=value, color=var, label=scales::label_percent(accuracy = 0.1)(value))) +
```
nmpeterson, line 55: Why this change?

dlcomeaux: I believe that @matthewstern made that change - he mentioned something was flagging as an error when using devtools::check() that was unrelated to this branch.

matthewstern: @nmpeterson the error here I think ended up being that `var` is not the number but the variable specifier so the `100*var` argument was producing a non-numeric error that was causing a halt in `check()`. I am quite confused as to why this showed up now, as opposed to earlier--my only guess is something having to with a package update or R 4.0?

Anyway, I had thought there was a package issue with sprintf, and you had used the `scales::percent` functions elsewhere, which is why I tried changing that first. I left the change because it felt more legible (esp. as an example) than the `sprintf`. 

If any of this is not what you had been going for, I'd be happy to revert!

nmpeterson: I'm going to submit a patch to revert this to `sprintf("%.1%%", 100*value)` for brevity. I want to avoid the double-parentheses syntax required to make `scales::label_percent` work, as I think it could cause confusion for a lot of users, and the label formatting is not the point of this sample code.

nmpeterson: Done: efe7d4273ad568193298e1aa2cfcd84ee5f5cf0f

PR #49, opened 2020-09-04
Transport samples

dlcomeaux:
I realized there was a discrepancy in the units of the data for transit ridership vs. what was described in the documentation (I had noticed this before uploading and thought I had fixed it, but apparently I didn't). No other changes.


dlcomeaux:
@matthewstern note - this is the change to the data that I was chatting with you about.

PR #52, opened 2020-09-17
Add global variables for font sizes

dlcomeaux:
Adds new definition for global font sizes in `cmapplot.R` and updates to include in `theme_cmap.R`


matthewstern:
We may want to use the list method for fonts as well, for consistency?

Issue #56, opened 2020-09-29
create debug_grobs mode in finalize_plot


matthewstern:
add an argument called `debug_grobs` or similar, and implement code that draws outlines around the various grobs and viewports if the argument is set to TRUE. 

With texbox grobs, this is probably best done in `box_gp`. In the plot, easiest is likely adding a theme element to the plot that paints a border. (this might change depending on @dlcomeaux 's legend implementation).

PR #58, opened 2020-09-29
legend justification, debug mode, and other improvements to theme_cmap() and finalize_plot()

dlcomeaux:
This PR changes legend alignment so that legends are now left aligned, with the option for indentation as desired. It uses gtables and the `arrange` function from `gridExtra`. I've also added the debugging function (or at least a start at it) that Matt referenced in #56.

Happy to talk more about what I did and the challenges I encountered, if helpful.


dlcomeaux:
Note that this is marked as a WIP because @matthewstern and I thought it would make sense to have `finalize_plot` exported first and finish out #53. This PR is up to date with the current status of that branch, and so it shouldn't be difficult to merge it into the master directly once #53 is done.

R/finalize_plot.R
```diff
@@ -201,18 +219,23 @@ finalize_plot <- function(input_plot = NULL,
     clip = "on"
   )
 
+  # extract height of legend object within ggplot plot
+  legend_height <- grid::convertUnit(ggpubr::get_legend(processed_plot)$heights[[3]],
```
matthewstern, line 223: Can we move this up into the initialization section?

R/finalize_plot.R
```diff
@@ -201,18 +219,23 @@ finalize_plot <- function(input_plot = NULL,
...
     just = c(0,0),
-    # size is function of remaining size available
+    default.units = "bigpts",
+    height = plot_constants$height - plot_constants$margin_v1_v2 -
+      plot_constants$margin_v4,
     width = plot_constants$width - plot_constants$title_width - plot_constants$margin_h3,
-    height = plot_constants$height - plot_constants$margin_v1_v2 - plot_constants$margin_v4,
-    clip = "on"
+    clip = "on",
+    gp = gpar(col = "black")
```
matthewstern, line 238: What's the purpose of this gpar function?

R/finalize_plot.R
```diff
@@ -315,29 +338,113 @@ finalize_plot <- function(input_plot = NULL,
                     col = cmapplot_globals$colors$blackish)
   )
 
+
+  # Function to create plot object with left aligned legend on top
+  buildChart <- function(p,l_margins,p_margins,l_indent,l_height,p_height) {
```
matthewstern, line 343: This function should be outside of finalize_plot() so that it gets built once by the package, as opposed to recreated each time someone runs finalize. In terms of documentation, see how Noel documented the ggplot_size_conversion function over in cmapplot.R.

R/finalize_plot.R
```diff
@@ -315,29 +338,113 @@ finalize_plot <- function(input_plot = NULL,
...
+                                                             r = updated_l_margins[2],
+                                                             b = updated_l_margins[3],
+                                                             l = updated_l_margins[4],
+                                                             "bigpts")
+                                      )
+                            )
+      )
+
+    # Extract the legend
+    l <- gtable::gtable_filter(gt, "guide-box")[[1]]
```
matthewstern, line 365: I'm a little bit confused here. if ggpubr::get_legend can extract a legend from a ggplot, why the reliance here on ggplot_build, gtable_filter, etc? I haven't touched the code at all, but it seems like get_legend could [do the extraction](https://rpkgs.datanovia.com/ggpubr/reference/get_legend.html) with more simply?

R/finalize_plot.R
```diff
@@ -315,29 +338,113 @@ finalize_plot <- function(input_plot = NULL,
...
+
+    # Remove the legend from the plot
+    p <- p + ggplot2::theme(legend.position = "none")
```
matthewstern, line 368: Could we use slightly less abstracted variable naming than "p", "l", "built", etc?

R/finalize_plot.R
```diff
@@ -315,29 +338,113 @@ finalize_plot <- function(input_plot = NULL,
...
+
+    # Create a buffer
+    buffer <- grid::rectGrob(gp = grid::gpar(col = "transparent", fill = "transparent"))
+
+    # Assemble
+    built <- gridExtra::grid.arrange(
+      gridExtra::arrangeGrob(
```
matthewstern: I believe that `grid.arrange` and `arrangeGrob` are the same thing, except that grid.arrange draws the figure and arrangeGrob just assembles it. Are you sure this doesn't work equally well without `grid.arrange()`?

R/finalize_plot.R
```diff
@@ -315,29 +338,113 @@ finalize_plot <- function(input_plot = NULL,
...
-        # modify text sizing. **THIS MODIFICATION IS NEEDED BUT NOT UNDERSTOOD**
-        text = ggplot2::element_text(size = cmapplot_globals$font$main$size * 1.25)
-      )
-    ),
+    p_l_stack,
     vp = vp.plot,
     name = "plot"
   )
 
+  # If debug is called, assemble additional outlines for overlay
```
matthewstern, line 404: I'd lean towards drawing outlines around the actual objects, as opposed to assembling additional objects that serve as outlines. In a buggy environment, there's a chance that your debug_stack won't actually align with where grobs and viewports are finding themselves placed.

R/finalize_plot.R
```diff
@@ -315,29 +338,113 @@ finalize_plot <- function(input_plot = NULL,
...
   final_plot <- grid::grobTree(
     grob_background, grob_topline, grob_title, grob_caption, grob_plot,
     name = "final_plot"
-  )
+    )
+  } else {
+    # If debugging, add outlines to text boxes and call debug outlines for plot
+    final_plot <- grid::grobTree(
+      grob_background, grob_topline,
+      grid::editGrob(grob_title, box_gp = grid::gpar(col = "black")),
```
matthewstern, line 440: Would a preferred way of doing this be to do something like `if(debug){ grob_outline_color = "black" } else { grob_outline_color = "transparent" }` way up top, and then adding `box_gp = gpar(col =  grob_outline_color)` in all textbox_grobs in all situations?

matthewstern (commented):
I know we're planning to dig into this more after the other PR is merged in, but I did have a few reactions that could be considered in advance of that.

R/finalize_plot.R
dlcomeaux: Yes I noticed that and addressed it already.

R/finalize_plot.R
dlcomeaux: If you know a good way to do this, that would be great. I had a great deal of trouble trying to add outlines to the objects in the ggtable, but that could just be unfamiliarity on my part. It didn't want me to add boundaries. Also, you can't add a border to a ggplot generally - it just lets you add a border around the panel, but not the labels.

R/finalize_plot.R
dlcomeaux: Makes sense.

R/finalize_plot.R
dlcomeaux: Yes, sure. Done.

R/finalize_plot.R
dlcomeaux: Oops, that's a holdover from a debugging attempt. Deleted.

R/finalize_plot.R
matthewstern: Or, better, maybe `box_gp = gpar(col = if(debug){"tomato1"}else{"transparent"})` where each grob gets a different color outline (yes, tomato1 is a [real color](http://sape.inf.usi.ch/quick-reference/ggplot2/colour)). 

I'd experiment with this first to make sure `gpar(col = "transparent"}` works... maybe it prefers NULL?

R/finalize_plot.R
matthewstern: Not off the top of my head but I'd be happy to look into it later this week.

R/finalize_plot.R
dlcomeaux: Done

R/finalize_plot.R
dlcomeaux: Sure.

R/finalize_plot.R
dlcomeaux: I had used that method because it was what that one you sent had, and it worked (!), but yes, it turns out that you can eliminate a couple calls and make it a tad simpler (although still complicated) with the get_legend function. I've adjusted.

dlcomeaux: And, doing it the new way should improve some debugging functions, which is a nice side benefit.

R/finalize_plot.R
dlcomeaux: Incorporated.

R/finalize_plot.R
dlcomeaux: I tried adding a border again with the new method, but it remains stubbornly not showing up. I've left the overlays in for now, as I don't have any better ideas at the moment.

R/finalize_plot.R
dlcomeaux: Also - `transparent` does work. I tried out a few things.

I'm going to leave it as all black for now, different colors for each would make the code much more complex (and I like the simplicity of the new approach).

R/finalize_plot.R
matthewstern: I've addressed this via borders in the ggplot theme.

matthewstern:
Didn't quite get as far as I wanted to over the weekend on this. Here's a list of some of the remaining things on my mind I'm interested in pursuing in this PR before we mark it ready -- these are not "must dos" but ideas for exploration and consideration:

- [ ] margin_v4 represents topline<->legend vertical spacing in finalize_plot. I'd like to implement this in theme_cmap so it is used to pull the legend away from the top of the plot window before finalize_plot is applied. Additionally, margin_v5 represents legend<->plot vertical spacing in finalize_plot, but only in `legend_build = adjust` mode. I believe there is a spacing attribute that can be passed to `ggplot2::theme()` that defines the distance between the legend and plot. We should implement this margin as that spacing, so that pre-finalize_plot, and post-finalize_plot but with `legend_build = safe`, this margin is applied.
- [ ] consider labeling or differentiating some way in cmapplot_globals between plot constants that are applied to theme_cmap vs to finalize_plot vs to both. Also, *consider* implementing an overrides structure into theme_cmap that mirrors what is used in finalize_plot.
- [ ] explore simplification/refinement of helper function. Some specific things that have come up--we currently use ggpubr::get_legend twice--once to measure the legend height and again to extract it. Is it feasible to run get_legend once to extract the legend, then measure the extracted legend? Also, in a situation where multiple aesthetics are mapped, ggplot will have multiple legends. This currently does not visualize correctly in finalize_plot because, I believe, the current height measurement extracts only the height of the first legend, not the cumulative height of all legends (the legend box, I believe).
- [ ] document cmapplot_globals in an Rd file
- [ ] update the margins explanation png with a debug = TRUE sample export.

matthewstern:
@dlcomeaux the current settings do seem to look right to me. Not sure where I got the -6 from. 

In other news, this is the plot I've been using to test the multiple-aesthetic/multiple-legend problem described above.

```
econ_plot2 <- ggplot(data = cluster_jobchange,
                    mapping = aes(
                      x = reorder(name, jobchange),
                      y = jobchange,
                      fill = category, 
                      alpha = assessment)) +
  geom_col() +
  coord_flip() +
  theme_cmap(gridlines = "v", hline = 0) +
  scale_y_continuous(labels = scales::comma)
```

R/finalize_plot.R
```diff
@@ -434,3 +463,75 @@ finalize_plot <- function(input_plot = NULL,
...
+  if(legend_build == "none"){
+    output_plot <- input_plot + theme(legend.position = "none")
+    return(ggplotGrob(output_plot))
+  }
+
+  # Reformat plot
+  ## IS ANY OF THIS ACTUALLY NECESSARY? CAN WE MOVE MORE OF THIS OVER TO THEME_CMAP?
+  format_plot <- input_plot + theme(
+    # ensure legend is left aligned
+    legend.position = "left",
```
matthewstern, line 490: This is needed to force multiple legends to stack vertically but is a fairly implicit/indirect way to do so. We should look into whether it is possible to do this more directly. If possible, this should be done at the `theme_cmap()` level so that it applies not only in `legend_build = "adjust"` mode. If it is possible, we should look into easy opportunities for customization.

R/finalize_plot.R
```diff
@@ -434,3 +463,75 @@ finalize_plot <- function(input_plot = NULL,
...
+    # update margins to account for possible indent
+    plot.margin = grid::unit(plot_constants$padding_plot,"bigpts"),
+    legend.margin = margin(t = plot_constants$padding_legend[1],
+                           r = plot_constants$padding_legend[2],
+                           b = plot_constants$padding_legend[3],
+                           l = plot_constants$padding_legend[4] + plot_constants$legend_indent,
+                           "bigpts")
```
matthewstern, line 497: Could plot.margin and legend.margin including legend_indent be applied to theme_cmap? Are they already, to some degree? It would be nice to remove this code from here if possible.

matthewstern: @dlcomeaux has pointed out that it is important to allow the user to adjust these items at the finalize_plot stage.

R/finalize_plot.R
```diff
@@ -434,3 +463,75 @@ finalize_plot <- function(input_plot = NULL,
...
+  legend_height <- grid::convertUnit(sum(legend$heights[3:(legend_total - 2)]),
+                                     "bigpts",
+                                     valueOnly = TRUE)
+
+  plot_height <- plot_constants$plotbox_height - legend_height - plot_constants$margin_v5
+
+  # Assemble a combined grob
+  built <- gridExtra::arrangeGrob(
+    # the legend, as a grob
+    grob(legend)[[1]],
```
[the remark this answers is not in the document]
matthewstern: I don't think it needs it. I've removed as of commit 7cec129

R/finalize_plot.R
matthewstern: added legend.box.just = "left" to theme_cmap in commit 7cec129. This seems to have fixed the problem

matthewstern:
@dlcomeaux I'm bleary eyed but I think I'm happy with where this is at. Curious for your thoughts? I've marked this as ready for review. Once you've taken a look and made any additional tweaks to my recent commits, I'd suggest adding at least Noel as a mandatory reviewer on this.

If you feel up for making a new margins explanation png, that would be swell, but is definitely not necessary right this second.

R/finalize_plot.R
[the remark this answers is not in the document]
dlcomeaux: Ok, if it works without it, great. I remember having trouble getting the legend to appear and when I changed this code it worked, but that could have been the result of some other changes as well.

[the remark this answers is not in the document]
matthewstern: <<shrug>>. Lets see if any problems show up in further testing--reverting to this might be needed. But, as I understand it, a tablegrob is a grob and should be able to be plotted without additional conversion.

dlcomeaux:
@matthewstern thanks, I gave this a brief review and made some tweaks (documented in the commit - I have a suggested way to disentangle 'margin_v3' from the top and bottom cases). I will try to make that margins png tomorrow, as well as add an additional example (since there are so many customization options and I think it would be useful to show them - maybe a variant of an existing one to demonstrate the options). I also want to investigate some plot margin behavior on the bottom, and do additional testing, which I will try out tomorrow.

nmpeterson:
>     * document cmapplot_globals (or at least and more specifically the plot_constants list) in an Rd file
>       **EDIT 1:** This is a larger conversation, because it looks like you can only document exported objects, and the environment isn't technically exported. One option is to make plot_constants an exported piece of data, rather than a part of the cmapplot_globals enviro. Curious for @nmpeterson 's input on this one. No matter what, it's not necessary for this PR.
>       **EDIT 2:** after sleeping on this, it might be entirely sufficient to just document the the plot_constants that can be modified at each stage (theme_cmap and finalize_plot) in the man files for those functions. For example, we could add a [custom section and bulleted list](https://r-pkgs.org/man.html#text-formatting) to each function's roxygen documentation.

I don't think plot_constants should be exported. I think a vignette documenting `cmapplot_globals` would be sufficient -- as would pushing the documentation of specific variables into the relevant exported functions.

matthewstern:
@dlcomeaux I pushed two quick commits this morning. I set up new documentation in theme_cmap() for the available overrides. Do you think that method seems reasonable/best? If so, and you have some time, could you implement the same for finalize_plot()?

As for the margins, I think disentangling margin_v3 is totally reasonable. Honestly, I'm frustrated by the vertical margins, because there are so many and they feel so unintuitively described. Do you have thoughts/feelings about how we can improve the situation? I am definitely not wedded to anything I've done up to now on this topic. There are a few thoughts on my mind, that don't necessarily agree with one another:

- if we are going to keep them numbered, there should be a logical flow (vertical or horizontal) to them. That means, unfortunately, that the margin you just created should be margin_v4 rather than margin_v6.
- another way to go about this, now that there are an even number in both columns, would be to use an excel-type "cell reference"-- margin_v2a would be over the title, with margin v2b over the legend, for example.
- for ease of use, should we also have margins that stretch across both columns? For example, using the above lingo, margin_v2 adds to v2a and v2b, but v2a and v2b are also available for changing values in one col or the other?
- should we go back to explicit naming, rather than numbers? say, margin_abovetitle, margin_abovecaption, margin_belowplot?

dlcomeaux:
@matthewstern my thought for renaming is that we do go back to descriptions, but with some abbreviated versions. I think we should keep margins separated, vs. having ones that can be multiple, as there are already so many that I think it would be more confusing than the value of the benefit it provides.

Here's a proposal for renaming:
```
margin_v1 = margin_topline_t
margin_v2 = margin_title_t
margin_v3 = margin_title_b
margin_v4 = margin_legend_t
margin_v5 = margin_legend_b
margin_v6 = margin_caption_b
margin_v7 = margin_plot_b
margin_h1 = margin_title_l
margin_h2 = margin_title_r
margin_h3 = margin_plot_r
```

matthewstern:
@dlcomeaux that sounds great. I'm supportive.

R/finalize_plot.R
```diff
@@ -159,26 +159,26 @@ finalize_plot <- function(plot = NULL,
...
-  plot_constants <- append(
-    plot_constants,
+  consts <- append(
+    consts,
     list(
       height = convertUnit(unit(height, "in"), "bigpts", valueOnly = TRUE),
       width = convertUnit(unit(width, "in"), "bigpts", valueOnly = TRUE),
       title_width = convertUnit(unit(title_width, "in"), "bigpts", valueOnly = TRUE),
-      margin_v1_v2 = plot_constants$margin_v1 + plot_constants$margin_v2
+      margin_topline_tb = consts$margin_topline_t + consts$margin_title_t
```
matthewstern, line 221: second item in this list should be _b, no?

R/finalize_plot.R
```diff
@@ -321,19 +321,19 @@ finalize_plot <- function(plot = NULL,
   # set caption textbox alignment options
   if(caption_valign == "top"){
```
matthewstern, line 372: with the improvements you've implemented to margins, specifically specifying margin_title_b and margin_caption_b separately, I think we can dramatically simplify this by setting the height of the grob to `grid::unit(consts$height - consts$margin_topline_tb - consts$margin_title_b - consts$margin_caption_b, "bigpts") - grid::grobHeight(grob_title)` and alternating only the alignment of the text (top or bottom" based on the `caption_valign` arg. This would eliminate the need for this captionvars list all together (as the only variable that would change based on the two settings is now `caption_valign`).

R/finalize_plot.R
matthewstern: oh, wait, nevermind. I misread your new taxonomy. sorry!

dlcomeaux: yes, good catch

dlcomeaux: ah yes, right, that was actually what was intended. I misread it when relooking haha

matthewstern:
@dlcomeaux, as of commit aa483d0 I reverted the right hand legend padding to -5, which seems to now look correct again for me in a variety of test plots, including this one:
test

Curious if you are seeing the same?

I'm done with this for now--I feel like it's in great shape. Barring any additional tweaks from Daniel, I think this is ready for merge after @nmpeterson 's review.

matthewstern:
@dlcomeaux your review and confirmation that I got everything right on the new [visual legend descriptor image](https://raw.githubusercontent.com/CMAP-REPOS/cmapplot/finalize_plot_legend_justification/man/figures/margins.png) would be appreciated

dlcomeaux:
> @dlcomeaux, as of commit [aa483d0](https://github.com/CMAP-REPOS/cmapplot/commit/aa483d07a70169b1199d31aea86e5a08c21e57d2) I reverted the right hand legend padding to -5, which seems to now look correct again for me in a variety of test plots, including this one:

Unfortunately, when I test it on a different computer, -5 is not enough, while -9 works. So this appears to be a system-specific (and/or an R-installation specific) issue. I'm not sure what would be the cause. A possible, if undesirable, workaround is to leave it as one of those two values and make a note that final tweaking might be needed using debug depending on system configuration.

dlcomeaux:
> 
> 
> @dlcomeaux your review and confirmation that I got everything right on the new [visual legend descriptor image](https://raw.githubusercontent.com/CMAP-REPOS/cmapplot/finalize_plot_legend_justification/man/figures/margins.png) would be appreciated

@matthewstern yes all of this looks right. I like `legend_bump`.

dlcomeaux:
So the main remaining issue is the exact alignment of legends. I did some investigating, played with several different margins (the legend box, the legend, etc.) and can't figure out what would be causing the discrepancy between your machine and mine, @matthewstern. While it's not ideal, in the interests of keeping this moving, the best solution might be to leave it as a "close enough" and then flag in documentation that this may need manual tweaking using `legend_bump`. Thoughts?

dlcomeaux:
Also as a heads up the commit above adds a new default legend key size (14, to be consistent with text sizing) and an additional element in `consts` that allows for easier modification of that for those unfamiliar with `legend.key.size` as a `theme()` call.

matthewstern:
> So the main remaining issue is the exact alignment of legends. I did some investigating, played with several different margins (the legend box, the legend, etc.) and can't figure out what would be causing the discrepancy between your machine and mine, @matthewstern. While it's not ideal, in the interests of keeping this moving, the best solution might be to leave it as a "close enough" and then flag in documentation that this may need manual tweaking using `legend_bump`. Thoughts?

Eh. I agree we should just move forward. I'll be curious to see what additional testing on other machines reveals but if 9 or 9.5 looks better on yours I'm happy to revert to that for the PR. sorry for changing it, I assumed I had tweaked some other number that forced a change.

dlcomeaux:
To be honest, that's what is confusing to me - that you had it working at
9.5 before, and now it's not. I assume you didn't change any of the
ggplot-wide defaults? That's the only thing I can think of. But sure, we
can revert to 9.5 for the moment. I'll do that and tweak the documentation.

matthewstern:
I didn’t change any ggplot defaults—just some tweaks in theme_cmap.

Sent from mobile
________________________________

matthewstern:
OK @dlcomeaux I give up! I don't know what I was doing Friday but your current padding of -9.5 works great for me. I wonder if I got the -5 while accidentally using a legend_bump at the same time? Or I changed something but forgot to `load_all()`? Who knows. Anyway, Sorry for the mess, but let's stick with -9.5 unless others raise an issue.

R/cmapplot.R
```diff
@@ -118,21 +118,40 @@ ggplot_size_conversion <- function(value, type = "bigpts") {
...
+  margin_legend_t = 5,  # [ ,f] Margin between top line and plotbox
+  margin_legend_i = 5,  # [t,f] Vertical margin between legends (only applies
+                        #        to multilegend plots)
+  margin_legend_b = 10, # [t,f] Margin between legend and plot (within plotbox)
+  margin_plot_b = 5,    # [ ,f] Margin between plotbox and bottom edge of image
+  margin_title_l = 2,   # [ ,f] Margin between left edge of image and title/caption
+  margin_title_r = 10,  # [ ,f] Margin between title/caption and legend/plot
+                        #        (Both title_l and title_r are deducted from `titlewidth`)
+  margin_plot_r = 10,   # [ ,f] Margin between plotbox and right edge of image
+  padding_plot = c(0,10,0,0),
```
nmpeterson, line 144: I don't love how this removes the padding around plots in the RStudio "Plots" pane. Now my titles and axis text are all smashed up against the edge of the pane...

matthewstern: lines 149 and 165 in `theme_cmap.R` were designed to address this on the top and the bottom in a situation with no titles. The idea was to create temporary spacing so it didn't look terrible at the `theme_cmap()` stage--but for this spacing to be stripped away once situated into the plotbox viewport during finalize. 

I clearly didn't think through the possible situations here, especially re the inclusion of title text. I'm open for suggestions on how to move forward. Is there an easy way to implement padding on all four sides of the larger ggplot object (regardless of the presence of legends, titles, axis titles, etc), that could get stripped away in finalize?

nmpeterson:
I think my only other comment is that I'm not convinced the pptx file for the margins diagram should live in the package. I did add .pptx files to the .Rbuildignore file, at least, so it won't be installed as part of the package -- you'd have to clone the GitHub repo to get it.

nmpeterson:
Also, you may have noticed that I set up a couple GitHub "actions" (specified by the .yaml files in .github/workflows/). These are checks that GitHub runs automatically on specified triggers:

- `pkgdown.yaml` will run any time the master branch receives a new commit, and automates the rebuilding of the pkgdown website, so we don't have to manually run `pkgdown::deploy_to_branch()`. (I think I snuck this action in to master prior to this PR being created, but now it's been merged in here, too.) A potential issue with this is that I don't think we'll be able to give this action access to Whitney, and it currently builds from macOS anyway, so the pkgdown site images are all rendered with Arial.
- `check-standard.yaml` essentially runs `devtools::check()` on both Windows and Mac every time master or a PR to master gets a new commit -- so my recent commits here have triggered it. (I originally had it testing on Linux, too, but that particualr check was running very slowly and felt gratuitous.) This way, it doesn't matter if we forget to run `check()` ourselves -- GitHub will tell us!

Both of these are mostly copied from [r-lib/actions](https://github.com/r-lib/actions), by the same team that wrote `devtools`, `pkgdown`, `roxygen2` and other great packages. I added badges to the readme file, which will show if the tests are passing or failing based as of the most recent commit.

I know these are totally outside of the scope of this PR, but, uh... YOLO.

R/cmapplot.R
matthewstern: I believe I have addressed this in commit ca042c1. @nmpeterson let me know if this solves your concerns?

nmpeterson: Looks much better, thanks

nmpeterson (approved):
Everything looks good to me. I still think maybe margins.pptx should find a new home outside of the package, but I won't fight anyone about it.

matthewstern:
> 
> 
> Everything looks good to me. I still think maybe margins.pptx should find a new home outside of the package, but I won't fight anyone about it.

@dlcomeaux I'm going to merge, but am happy to discuss removing the pptx file in a later commit.

PR #59, opened 2020-09-30
Highlight color function

gritzenthaler:
I named this branch a long long time ago; no edits are actually made to Theme. 

The highlight color is currently blue, but this can easily be changed, as well as the non-highlighted color (gray).


R/highlight_function.R
```diff
@@ -0,0 +1,48 @@
...
+#'        aes(x = system, y = ridership, fill = system)) +
+#'    geom_col() +
+#'    cmap_fill_highlight(transit_ridership$system, "metra")
+#'
+#' @export
+cmap_fill_highlight <- function(variable, highlight) {
+
+  n <- length(unique(variable))
+
+  palette <- rep.int(c(other = "#9daab3"), n)
```
dlcomeaux, line 20: It seems like this palette color (and the highlight color below) could be useful to have as a functional argument? Defaulting to the ones you've chosen, but with the option to change if desired?

dlcomeaux:
There might be a way to turn this into just one function, actually - using the `last_plot()` function. In particular, if you call
`ggplot2::last_plot()$mapping$fill` on a plot that uses `col`, it will return `NULL`. We could use that to determine which branch to go down, in something that looks like this:

```
cmap_highlight <- function(variable, highlight) {

  n <- length(unique(variable))

  palette <- rep.int(c(other = "#9daab3"), n)

  names(palette) <- levels(factor(variable))

  palette[[highlight]] <- "#009ccc"

  if (is.null(ggplot2::last_plot()$mapping$fill)) {
  ggplot2::scale_col_manual(values = palette)
  } else {  
  ggplot2::scale_fill_manual(values = palette)
  }

}
```

This seems to work when I test it, but if others are more familiar with the specifics of how `last_plot` works (and whether there are any cases where it might have a mapping for both `fill` and `col` that would be good to think about).

gritzenthaler:
Oh ok, I see what you mean now. I guess if it's getting way too complicated we can leave the two functions as is. I can rework this right now to include an optional highlight color input with a default.

matthewstern:
Because this is meant to be an object added to a standard ggplot construction, and as Noel has encouraged us to keep with ggplot nomenclature as much as possible, distinct function names for fill and color (and colour, because why not) seem like the right way to frame this to me. 

However, because the code within the two functions is nearly identical, you could consider creating an internal function that exists only once and is called by the various other functions. Something like:

```
make_highlight_pallet <- function(variable, color1, color2){
  n <- length(unique(variable))
  palette <- rep.int(c(other = color2), n)
  names(palette) <- levels(factor(variable))
  palette[[highlight]] <- color1
  return(palette)
}
```

I'm not sure if a similar construction is worth considering in the other color functions as well?

gritzenthaler:
Adding an internal palette prep function was easy enough. There is now an option to change the highlight color.

R/highlight_function.R
```diff
@@ -0,0 +1,55 @@
...
+#' @export
+cmap_fill_highlight <- function(variable, highlight, color = "#009ccc") {
+
+  palette <- make_highlight_palette(variable, highlight, color)
+
+  ggplot2::scale_fill_manual(values = palette)
+
+}
+
+#' @rdname cmap_fill_highlight
```
matthewstern, line 43: I like the documenting all in one place. It [looks like](https://r-pkgs.org/man.html#multiple-man) `@describeIn` is the more standard way to do this. Did you pick `@rdname` for a reason?

R/highlight_function.R
```diff
@@ -0,0 +1,55 @@
+#' Highlight palette prep function
+#'
+#' Create named list to serve as palette input for following fill/color highlight functions
+#'
+make_highlight_palette <- function(variable, highlight, color = "#009ccc"){
```
matthewstern, line 5: I think arg defaults in the subfunction are unnecessary and potentially confusing, as the defaults are already set in the arguments of the top-level functions.

R/highlight_function.R
```diff
@@ -0,0 +1,55 @@
...
+#' @param color (optional) Specify the highlighted color as a hexcode string. Default is #009ccc (blue)
+#'
+#' @examples
+#' ggplot(dplyr::filter(transit_ridership, year=="2019"),
+#'        aes(x = system, y = ridership, fill = system)) +
+#'    geom_col() +
+#'    cmap_fill_highlight(transit_ridership$system, "metra")
+#'
+#' @export
+cmap_fill_highlight <- function(variable, highlight, color = "#009ccc") {
```
matthewstern, line 35: I think it's worth adding the un-highlighted color here too, so the user can overwrite it if need be.

R/highlight_function.R
```diff
@@ -0,0 +1,55 @@
+#' Highlight palette prep function
+#'
+#' Create named list to serve as palette input for following fill/color highlight functions
+#'
```
matthewstern, line 4: I believe this wants a `@noRd` tag, at least according to what I've seen @nmpeterson write.

matthewstern (commented):
@gritzenthaler I left a few comments on some specific lines of code, but in general i think this looks great! 

One question for you: it looks like your PR is attempting to overwrite a lot (maybe all) of the images from the readme, including at least a partial implementation of the modified y axis placement code (even though though I don't see that code in this PR). (See the [readme within the branch](https://github.com/CMAP-REPOS/cmapplot/tree/theme_gcr).) Did you by any chance reknit the readme at some point, and then remove that experimental code? Tangentially, the y axis placement stuff actually looks really good! Is that code around somewhere to look at?

gritzenthaler:
Thanks for the comments! I was using @rdname just because I didn't know @describeIn was a thing. 

As for the Theme y axis adjustments, I took them out for now since extending the ticks isn't consistent and looks kind of stupid for long axis labels. What do you think? Should we get that feature at least halfway by bumping the labels up, or leave them normal-looking for the instances where we'll use the theme without Comms making adjustments?

matthewstern:
@gritzenthaler I agree we should leave them out for now. In case we return to them, your code is a solid start. For posterity, it looks like you implemented it in `theme_cmap()` as of [this commit](https://github.com/CMAP-REPOS/cmapplot/pull/59/commits/5b2572f0aa8d01c684d22e4389132e9b183486af).

I added tweaked a few things, made the documentation a bit more verbose, and added the ability to specify multiple fields to highlight. I'm going to go ahead and merge it. Thanks for putting this together!

README.md
```diff
@@ -38,6 +40,7 @@ ggplot(cluster_jobchange, aes(x = reorder(name, jobchange), y = jobchange, fill
   geom_col() +
   coord_flip() +
   theme_cmap()
+#> 'Whitney' font family not found. Using a substitute...
```
nmpeterson, line 43: Just want to point this out. We should try to avoid re-knitting on machines without Whitney. Not a big deal right now, especially since none of us currently have access to Whitney, but just something to keep in mind.

PR #64, opened 2020-10-16
Midpoint setting

dlcomeaux:
Allows setting of midpoints on diverging palettes (defaults to 0). I've configured it so that it only applies to palettes that begin with `div`.


matthewstern:
With Greta's thumbs up I think this is ready to merge. @gritzenthaler, it does seem reasonable to allow users to extract discrete palettes from continuous palettes if they need and it's not too hard. This is now #70 .

PR #65, opened 2020-10-20
Recession tweaks

matthewstern:
This PR makes the recessions table internal, and creates a new external function called `update_recessions()` that users can use to fetch new recession data from NBER.

@nmpeterson, looking for a bit of an assist here on a few outstanding issues:
- the new function `update_recessions()` uses a number of packages not required elsewhere in the package--they really shouldn't be required, I would think. I added them to `Suggests:`  in DESCRIPTION but check() is still throwing the following error: `Namespace dependencies not required: 'readxl', 'stringr', 'tibble'`. What is the best way to handle this?
- How do you get `geom_recessions()` to accept the global environment's version of the table `recessions` when it exists (say, after running `recessions <- update_recessions()`), rather than using the internal version of the table? Would it make the most sense to add a argument to `geom_recessions`, say, `recession_src = NULL`, where `recession_src` is a globally available data table that overwrites the internal `recessions` data table if the user sets a non-Null value?
- Are there ways that we can simplify `update_recessions()` now that it is internal? One thing on my mind is whether we can eliminate the conversion to a tibble from a df and avoid the requirement of the `tibble` package?

Remaining to dos:
- [x] I have not yet modified the gh-pages files to account for the updates to what is and is not exported by the package. This should be updated within this PR before merging.


nmpeterson:
I think the only way to avoid the extra imports would be to rewrite `update_recessions()` so that it no longer uses them. (But I don't think those imports are that big of a deal anyway -- two of them are core tidyverse packages and therefore already imported if the user has run `library(tidyverse)`.)

How long does `update_recessions()` take? If not terribly long, maybe it should just be done in cmapplot.R as part of loading the package. Alternatively, maybe it could be called automatically when `geom_recessions()` is called.

matthewstern:
> How long does `update_recessions()` take? If not terribly long, maybe it should just be done in cmapplot.R as part of loading the package. Alternatively, maybe it could be called automatically when `geom_recessions()` is called.

@nmpeterson, not long. This could work. We can leave the existing `recessions` table as internal data, and add an argument `geom_recessions(update_recessions = TRUE)` that runs `update_recessions()` and stores the table as `recessions` within the function's environment. I think, then, that it will take the function environment's version of the table if it exists, and fall back to the package environment's table if the local one doesn't exist. Does that sound right?

nmpeterson:
Yeah, that sounds like a good way to do it. You'll also have to allow for the possibility that the file you're fetching from NBER is unavailable (either temporarily or permanently), and fall back on the internal data in that case, too. Would be good to print a warning in this case.

matthewstern:
This was a bit more complicated than I expected because creating a table `recessions` within the topline function doesn't automatically make it available to subfns due to how R handles function environments. However, I believe I have gotten it working but wouldn't mind a second set of testing eyes. Alternatively, I will try testing again on more sleep.

@nmpeterson I have yet to learn how to successfully use `try` or `trycatch`. I presume that's the best way to implement the fall-back you're recommending? Would you mind implementing this error handling if it's easy for you to do? The code in question is lines 136-138.

nmpeterson:
What's the R/sysdata.rda file?

matthewstern:
> 
> 
> What's the R/sysdata.rda file?

[Apparently](https://r-pkgs.org/data.html#data-sysdata), this is where R likes to store internal data.

PR #67, opened 2020-10-23
Correct fonts in labels

dlcomeaux:
Use finalize_plot() to update font faces, families, sizes, and colors for text added to the plot through geom_text() or annotate(). Requires hard-coded alternative to default value because font sizes are not playing nice.


matthewstern:
I'm OK with this change although (I know I sound like a broken record but...) I don't love that it kicks in at finalize and not at theme_cmap. I can't think of a clean or effective way to put it in theme, though. 

We should specify in documentation, I think, that font can be set by specifying aesthetics in geom_text() if the user wants to. Which brings back up the question of user access to variables in `cmapplot_globals`.

dlcomeaux:
@matthewstern I agree that doing this in finalize is not ideal. This _may_ be an issue of having an incomplete theme, although I'm not sure about that. That's my best guess, though.

To your question - after I use `load_all()` I am able to access things in the `cmapplot_globals` stack, so I think that they are user-callable. Unless there are things there that are usable with `load_all` but not if you install the package? I'm not sure how that works.

nmpeterson:
`cmapplot_globals` is _NOT_ available when cmapplot is loaded via `library(cmapplot)`.

dlcomeaux:
Oof. Ok, good to know.

matthewstern:
Following up on the discussion [here](https://github.com/CMAP-REPOS/cmapplot/commit/23f0182843c643116ed1ebd418a8c1fa7e325437), it looks like the authors of ggplot [recommend](https://ggplot2.tidyverse.org/articles/ggplot2-in-packages.html#creating-a-new-theme) explicitly building off a built-in ggplot theme if creating a custom theme. This should enable our theme to be 'complete'. Perhaps this in combination with an understanding of the `base` theme arguments might help us solve our font size issue. A man can dream...

I can try to look into this in the coming days if no one else gets to it first.

R/finalize_plot.R
```diff
@@ -261,6 +261,20 @@ finalize_plot <- function(plot = NULL,
...
+  default_fontface <- ggplot2::theme_get()$text$face
+  default_fontcolor <- ggplot2::theme_get()$text$colour
+  default_fontsize <- ggplot2::theme_get()$text$size
+
+  ggplot2::update_geom_defaults(
+    geom = "text",
+    new = list(family = cmapplot_globals$font$label$family,
+               face = cmapplot_globals$font$label$face,
+               size = cmapplot_globals$font$label$size,
+               colour = cmapplot_globals$colors$blackish)
```
nmpeterson, line 275: Can this be extended to apply to `geom_text_lastonly`?

R/cmapplot.R
```diff
@@ -44,7 +44,8 @@ if (.Platform$OS.type == "windows") {
       note = list(family="Whitney Book", face="plain", size=11),
       title = list(family="Whitney Semibold", face="plain", size=17),
       axis = list(family="Whitney Book", face="plain", size=14),
-      label = list(family="Whitney Semibold", face="plain", size=14)
+      label = list(family="Whitney Semibold", face="plain", size=5)
```
nmpeterson, line 47: These are going to need to be reconfigured to match @matthewstern's font restructuring once #72 is merged.

dlcomeaux: Yup, good flag.

R/finalize_plot.R
dlcomeaux: I think so? Some of this may no longer be necessary after we update to the complete theme, as I think that solves some of the inheritance problems...so we will have to redo some testing before doing anything with this branch.

dlcomeaux:
Changing to draft because we will need to give this some more thought after #72 is merged.

dlcomeaux:
I updated this to accommodate the new `theme_cmap` conventions. Unfortunately, the font sizes still don't correspond to what we would want them to - you need to input a size of 5 for it to output at size 14, which I've done by multiplying `M` by `5/14`. This is certainly not ideal, and I'm open to other ideas on what might be causing the problem. For now, though, this works.

nmpeterson:
If the labels are supposed to use Whitney Semibold (which you had in your previous iteration), then they should now use the `strong` font spec instead of `regular`.

dlcomeaux:
Oops, good catch, yes @nmpeterson

matthewstern:
I presume you've seen this but it looks like you're not the only one who has decided 14/5 is the right ratio. 

https://stackoverflow.com/questions/25061822/ggplot-geom-text-font-size-control/25062509

"it's the difference in units, geom_text default of 5 might be 5mm and the theme() size unit is point. 1 point is 1/72 inch=0.35mm, so 1 in geom_text() is 1mm, 1/0.35 =~ 14/5"

R/finalize_plot.R
matthewstern: I believe we can set `geom_text_lastonly` defaults in the `default_aes` section of the `GeomTextLast` ggproto object created in geom_text_lastonly.R. I haven't experimented with whether `update_geom_defaults()` works with custom ggproto objects, although I would assume it does not.

nmpeterson: @matthewstern if I remember our earlier conversations correctly, I had proposed modifying the function to use the CMAP font spec by default, but we decided not to implement that because it would look weird when `theme_cmap` was not also being used.

nmpeterson:
@dlcomeaux @matthewstern I just discovered that `ggplot2::.pt` (=2.845276, slightly greater than 14/5) is used internally by `ggplot2` to convert mm to pt and vice versa. Help description for `.pt` and `.stroke`: "Multiply size in mm by these constants in order to convert to the units that grid uses internally for `lwd` and `fontsize`."

R/finalize_plot.R
nmpeterson: @matthewstern @dlcomeaux one more note on this: if we export the `GeomTextLast` ggproto object (as ggplot2 does for its own ggproto objects) then we should be able to use `update_geom_defaults("TextLast", ...)` on it.

matthewstern:
> 
> 
> @dlcomeaux @matthewstern I just discovered that `ggplot2::.pt` (=2.845276, slightly greater than 14/5) is used internally by `ggplot2` to convert mm to pt and vice versa. Help description for `.pt` and `.stroke`: "Multiply size in mm by these constants in order to convert to the units that grid uses internally for `lwd` and `fontsize`."

This suggests that we may want to revisit our size conversion function. It may become irrelevant if we can directly convert sizes to mm and then multiply explicitly by ggplot's built-in constant. The help file isn't clear... I presume .stroke is for lwd and .pt is for font size? Interesting that they are different from one another.

nmpeterson:
That was the impression I got, too, but yes -- unclear.

dlcomeaux:
OK @matthewstern @nmpeterson I'll give it a try and see if I can make a new version of the function work.

dlcomeaux:
Apparently `.stroke` doesn't really have much use (per [here](https://github.com/tidyverse/ggplot2/issues/3334))

matthewstern:
[This search](https://github.com/tidyverse/ggplot2/search?q=repo%3Atidyverse%2Fggplot2+%22.pt%22) shows you where `.pt` is used, and pulls up this text from one of the ggplot vignettes:

The `size` of text is measured in mm. This is unusual, but makes the size of text consistent with the size of lines and points. Typically you specify font size using points (or pt for short), where 1 pt = 0.35mm. ggplot2 provides this conversion factor in the variable `.pt`, so if you want to draw 12pt text, set `size = 12 / .pt`.

Per [this search](https://github.com/tidyverse/ggplot2/search?p=2&q=repo%3Atidyverse%2Fggplot2+extension%3Ar+%22.stroke%22), it looks like `.stroke = 96 / 25.4` but that the constant is unused in the package.

dlcomeaux:
I tweaked the function to include both `.stroke` and `.pt` because it seemed a bit more elegant than the random adjustments we had before. However, it's arbitrary, so if folks think it is too cute by half, we can adjust again. Line width conversions do appear to be distinct from the text font size conversions, for whatever reason, so my solution there is just to hardcode a division by `.pt` in the relevant call in finalize. We could make a separate function (or make this one more complicated) but given that it is a one-use thing, and the division is pretty simple, I propose keeping it, if we want to keep the adjustment in finalize. I am open to ideas, however, if anyone has a clue how we could include this in `theme_cmap` without having to have the user manually specify it.

matthewstern:
Based on the following sample code, I am starting to think that the 96/72.27 conversion factor applied in `ggplot_size_conversion()` (explicitly before, implicitly now) is actually not necessary. In the following, the first ggplot looks correct to me, while the second does not:

```
library(tidyverse)
library(grid)

size_pts <- 100
size_mm <- size_pts/.pt

# confirm /.pt is the same as points to mm:
all.equal(size_mm, convertUnit(unit(size_pts, "pt"), "mm", valueOnly = TRUE))
all.equal(size_mm, size_pts * 25.4/72.27)

ggplot_size_conversion <- function(value, type = "bigpts",isFont) {
  # convert input type to mm (if not already)
  value_in_mm <- grid::convertUnit(grid::unit(value, type), "mm", valueOnly = TRUE)
  
  return(
    value_in_mm
    * ggplot2::.stroke # Multiply by units for R pixels per mm (=96/25.4)
    / ggplot2::.pt     # Account for the ggplot2::.pt factor (=72.27/25.4)
  )
}


# this looks right 
ggplot(data = tibble(x = 1, y = 0), aes(x = x, y = y)) +
  geom_hline(size = size_mm, yintercept = 0, color = "grey90") +
  geom_text(label = "A line! ggg", colour = "red", size = size_mm) +
  theme_minimal()

# this doesn't look right (to me)
ggplot(data = tibble(x = 1, y = 0), aes(x = x, y = y)) +
  geom_hline(size = ggplot_size_conversion(size_pts, type = "pt"), yintercept = 0, color = "grey90") +
  geom_text(label = "A line! ggg", colour = "red", size = size_mm) +
  theme_minimal()
```

Am I missing something? If not, I would lean towards removing the function `ggplot_size_conversion()` entirely, as it is only utilized in a single place in `cmapplot`. The user can either use `/ggplot::.pt` for points or `grid::convertUnit()` for any other unit.

On a related note, it's worth noting here that we are inconsistent in using bigpts and pts in the package as of now. We use bigpts for most things, but fall back to points when it comes to text size and the legend key (which we want to be equal to the legend text). I would think we should try to be consistent? Seems to me that Adobe points/Bigpoints/[desktop publishing points](https://en.wikipedia.org/wiki/Point_(typography)) of 1/72 inch is the way to go, although it does weird me out a bit that ggplot and grid use pt of 1/72.27 inch as their defaults. 

If we did this, it would negate using `/ggplot::.pt` for conversion to mm. We could use `grid::convertUnit()` or define our own `.bigpt = 72/25.4`.

dlcomeaux:
@matthewstern I just ran both of those, and the reason I would advocate we keep the size conversion function is that while those line up nicely with the text, they are **not** the actual width you would expect. The line that goes through `ggplot_size_conversion` is 1.38 inches, as expected (100 pts, times .0138 inches per point). The line that is just 100 pts divided by `.pt` is ~1.03 inches, which is what you would expect (it is the same, minus the scaling factor of 96/72.27).

I think we can/should migrate to using `bigpts` throughout, though, regardless of this decision.

matthewstern:
OK, based on @dlcomeaux's point about the measured line size and Ben K's statement that the following graph yields the most accurate definition of the line and font being the "same size", I'm on board with the current conversion and ambivalent about doing the conversion in a stand-alone function vs in line. 

We do need to consider how to replace `.pt` if we're going to stick to bigpts throughout. 

```
ggplot(data = tibble(x = 1, y = 0), aes(x = x, y = y)) +
  geom_hline(size = ggplot_size_conversion(size_pts, type = "pt"), yintercept = -0.06, color = "grey90") +
  geom_text(label = "A line! ggg", colour = "red", size = size_mm) +
  theme_minimal() +
  scale_y_continuous(limits = c(-1,1))
```

R/finalize_plot.R
matthewstern: In commit 09584be I have exported all our proto objects and created dummy documentation for them. I can confirm that `update_geom_defaults` does work on our custom geoms. 

Test code:
```
  geom_text() +
  geom_text_lastonly()

g # all text black

update_geom_defaults("Text", list(color = "blue"))
update_geom_defaults("TextLast", list(color = "red"))

g # text now blue and red
```

matthewstern:
I believe this is ready for review. In addition to incorporating the changes from #81, this PR also:

- addresses sizing concerns by establishing a new constant `.lwd` and updating the function now called `gg_lwd_convert`. 
- corrected the topline width drawn by finalize
- renames the constant `lwd_originline` to `lwd_strongline` as its use is not restricted to origins
- many improvements to the `plots.Rmd` vignette, including incorporating a new section on the new `apply_cmap_default_aes`

R/default_aes.R
```diff
@@ -0,0 +1,174 @@
...
+#'             aes(x = year, y = realgrp, color = cluster)) +
+#'   geom_recessions(ymax = 0.4, text_nudge_x = 0.1) +
+#'   theme_cmap(hline = 0,
+#'              axislines = "x",
+#'              legend.max.columns = 2) +
+#'   ggtitle("Change in gross regional product over time") +
+#'   geom_line() +
+#'   scale_x_continuous("Year", breaks = seq(2007, 2017, 2)) +
+#'   coord_cartesian(clip = "off") +
+#' geom_text_lastonly(aes(label = realgrp), add_points = TRUE)
```
nmpeterson, line 72: Are we happy with the default point size for `geom_text_lastonly` after applying `finalize_plot` or `apply_cmap_default_aes`, or should they be bigger? They look a little small to me with the thicker lines, but I'm not sure if we have guidance from Comms on this?

vignettes/plots.Rmd
```diff
@@ -19,16 +19,14 @@ knitr::opts_chunk$set(
 
 library(tidyverse)
 library(cmapplot)
+unapply_cmap_default_aes()
```
nmpeterson, line 22: I don't understand why `unapply` is needed here.

matthewstern: This was necessary for testing purposes. Without it, repeat knitting of the vignette within an R session causes all the early plots to be drawn with cmap aesthetics. Once the vignette is done being developed, it is no longer useful. However, because this function simply puts back the defaults that were just extracted in the previous line of code, it strikes me as pretty harmless.

R/default_aes.R
matthewstern: Good point. We have models for this

nmpeterson: I set the default size to 2x the default line width (so, 6 pt?) in commit 2a48ec4ba8ba9a6496322d5488249b2508a6d73e. Looks good enough to me, but we may want to check with Ben at some point.

vignettes/plots.Rmd
nmpeterson: Ok, just wanted to make sure it wasn't doing something critical that I was unaware of.

R/geom_text_lastonly.R
```diff
@@ -128,13 +128,12 @@ geom_text_lastonly <- function(mapping = NULL, data = NULL,
 #' The \code{cmapplot} package contains a few custom ggproto objects. For the
 #' most part, these are slightly tweaked versions of ggplot2's default proto
 #' objects. For more information about these, see
-#' \code{\link[ggplot2]{ggplot2-ggproto}}.
+#' \code{\link[ggplot2::ggproto]{ggplot2::ggproto}}.
```
matthewstern, line 131: @nmpeterson this change is now causing a warning for me in `check()`:
```
Missing link or links in documentation object 'customproto.Rd':
    '[ggplot2::ggproto]{ggplot2::ggproto}'
```

What links do and don't work is totally beyond me (I haven't done any testing of links and when I load help files in R the links haven't been active) but I think this help page is titled ggplot2-ggproto?

image

nmpeterson: Ah, crap, I must have forgotten the `2` when I was trying to bring up the help, and concluded that you intended to link to `?ggroto` (which also exists, but is for creating new ggproto objects). I will revert this.

nmpeterson: Should be fixed by commit 2246b3e3d1e72de56bd9f2c849db65de36c5dbae

R/theme_cmap.R
```diff
@@ -275,7 +278,7 @@ theme_cmap_base <- function(consts = cmapplot_globals$consts,
     panel.background =   NULL,
     panel.border =       element_blank(),
     panel.grid =         element_blank(),
-    panel.spacing =      unit(consts$half_line, "pt"),
+    panel.spacing =      unit(consts$half_line * 2, "pt"),
```
dlcomeaux, line 281: Is there a reason we are using points here, and not big points? I understand why the legend key is being done in points, but I'm not sure about this one.

matthewstern: a reasonable argument. Technically, the `half_line` comes from the font size, which is why these were all in `pt`. But, it does seem reasonable to adjust them all to `bigpts`. I'll adjust now.

PR #72, opened 2020-10-29
Rethinking theme construction

matthewstern:
This started as a small inspection into fontsizes, but ended up getting much bigger. I'm going to open this as a draft so people can see where I'm going on this.


matthewstern:
Commits up through b092906 break the package overall, but are starting to build up some experimental changes.

- Our font construction matrix was getting complicated. All family/face/size combinations are called via individual components within theme and finalize, so I don't believe there's a need to specify every possible permutation like we were doing. As of now, the `cmapplot_globals$fsize` list contains three font sizes: 11, 14, and 17. The `cmapplot_globals$font` list still contains font family and face info (no longer size), but this is now just "strong", "regular", and "light". 
- The `check_cmap_fonts()` function is now the `display_cmap_fonts()` function. In response to the changes above, each display in this function now displays both the (consistent) varnames that can be called and the (variable) fonts in use on the local machine.
- Building off of lessons learned from too many ggplot2 help pages to count, I have constructed an actual complete theme called `theme_cmap_base` in the file `theme_cmap2.R`. This is modeled off of `ggplot2::theme_grey()`, which is, as far as I can tell, the only independently constructed complete theme anywhere. This theme is not necessarily perfect as of yet, but it's close--most of our default values are included. And inheritance seems to actually work now!
- FONTS AND SIZES: ah, yes, why I started this branch in the first place. Right now, when I make a graph with `theme_cmap_base`, and then save the output to PDF using Cairo graphics, **the font sizes work without any adjustment**. Some fonts are showing up as "CairoFont" so there's still some work to do there.
- the negative right margin for the legend is no longer necessary! @dlcomeaux you'll be interested to hear that it appears the area to the right of the left-most legend key is, actually, the space between the legend title (nonexistent in our theme) and the legend item. This appears to not be modifiable, but does appear to be exactly one lineheight. This was showing up as 9.5 because of the weird resizing issue before (which I think I now understand), so we can modify it directly in the theme by accessing the default lineheight from cmap_globals. 

next steps I plan to tackle:
- [ ] explore the CairoFont outputs and try to deal with them
- [ ] rebuild the modifiable theme_cmap() so it builds on the new theme_cmap_base()
- [ ] rebuild finalize with new font structure

Very curious for y'all's thoughts about/concerns with this direction.

nmpeterson:
I think this looks great. I like the font restructuring. A few thoughts:

- `theme_cmap_base()` should probably be folded into theme_cmap.R, rather than a separate file.
- Actually, is there a reason for `theme_cmap` to continue to exist? Or does `theme_cmap_base` completely replace it? (If so, it should take `theme_cmap`'s name.)
- I'd like to rename the font sizes to S/M/L, which would be easier to extend in the future if we need to (with XS, XXL, etc.) and also would remove the "regular"/"reg" overlap between the font name and size specifications.

matthewstern:
Thanks Noel!

Some quick responses:

> 
> `theme_cmap_base()` should probably be folded into theme_cmap.R, rather than a separate file.

Yes, this is my goal. It is still a work in progress.

> 
> Actually, is there a reason for `theme_cmap` to continue to exist? Or does `theme_cmap_base` completely replace it? (If so, it should take `theme_cmap`'s name.)

My plan was to have the base function be an internal theme with all possible attributes the user might want (axis lines, origin lines, etc). `theme_cmap` would call the base theme then selectively "turn off" elements. It may make sense for `theme_cmap` to also "turn on" elements in some situations. This is roughly how the ggthemes package works--see their `theme_foundation()`.

>
>    I'd like to rename the font sizes to `S`/`M`/`L`, which would be easier to extend in the future if we need to (with `XS`, `XXL`, etc.) and also would remove the `regular`/`reg` overlap between the font name and size specifications.

That makes sense, I can incorporate this easily.

> 
>    I think the "CairoFont" issue should be a very low priority for fixing, as long as the output visually appears to be using the correct typeface.

Fair enough.

nmpeterson:
Given the significant font spec restructuring, I think it probably makes sense to do the `.onLoad()` stuff in this branch instead of a separate one to avoid headaches. I'll push a commit for that later today.

dlcomeaux:
@matthewstern I think this is great. I agree with Noel's comments about relative priority. And I'm glad that we are **finally** starting to make some headway on these font sizing and inheritance problems! Great work.

To a specific question, I think it makes sense to maintain `theme_cmap_base` for the reasons you stated.

matthewstern:
OK, this is starting to look pretty decent. There is work to do bringing documentation up to date in theme_cmap(), finalize_plot(), and the now-exported cmap_globals. Additionally, the ppt helper visualization will need to be updated. Testing is welcome at this stage, but I'll try to clean up documentation and set this PR as ready to review soon.

R/cmapplot.R
```diff
@@ -132,44 +179,29 @@ ggplot_size_conversion <- function(value, type = "bigpts") {
...
 
-# Define common colors
-cmapplot_globals$colors <- list(
-  blackish = "#222222"
-)
+
+
+
+# temp plots ---------------------------------------------------
+# some plots without themes for internal testing
```
matthewstern, line 187: Erase these

R/finalize_plot.R
```diff
@@ -519,23 +503,57 @@ finalize_plot <- function(plot = NULL,
...
+  # add debug rects around legend if in debug mode
+  if(debug){
+    plot <- plot + ggplot2::theme(
+      legend.background = element_rect(color = "red"),
+      legend.box.background = element_rect(color = "red")
+    )
+  }
+
+  # Determine correct legend margins to use. Use from plot (set
+  # via theme_cmap(), with possible overrides, unless user has
```
matthewstern, line 539: Missing )

R/finalize_plot.R
```diff
@@ -554,8 +572,10 @@ buildChart <- function(plot,
   if (number_of_legends > 1) {
     # if multilegend plot, establish loop to change margins
     for (i in 1:(number_of_legends-1)) {
```
matthewstern, line 556: How does this react to a plot with no legend? I think there is a "step through" function that is preferable here

R/finalize_plot.R
```diff
@@ -519,23 +503,57 @@ finalize_plot <- function(plot = NULL,
 # Function to create plot object with left aligned legend on top
 buildChart <- function(plot,
                        consts,
-                       legend_build) {
+                       overrides,
+                       legend_build,
```
matthewstern, line 507: Let's change to legend_shift as logical and remove legend elimination from here--maybe add to theme?

R/finalize_plot.R
```diff
@@ -554,8 +572,10 @@ buildChart <- function(plot,
   if (number_of_legends > 1) {
     # if multilegend plot, establish loop to change margins
     for (i in 1:(number_of_legends-1)) {
-      margin_index <- 2*(i+1) # e.g., for a 2-legend item, this modifies element 4
-      legend$heights[[margin_index]] <- grid::unit(consts$margin_legend_i,"bigpts")
+      # e.g., for a 2-legend item, this modifies element 4
+      margin_index <- 2*(i+1)
+      # apply correct legend spacing
+      legend$heights[[margin_index]] <- grid::unit(margin_legend_i,"bigpts")
```
matthewstern, line 578: Presuming this is already set correctly barring an override, this whole section could be shifted up into the if override exists section

matthewstern:
This is ready to be evaluated. Here's an attempted summary of changes:

- cmapplot_globals now an exported list, with independent documentation
- fonts and font sizes now specified separately, with upgraded test function `display_cmap_fonts`.
- fonts should load upon actual loading of package via `library(cmapplot)`. @nmpeterson, `check()` warns about preferring messages in `.onAttach` rather than `.onLoad`-- just checking that you're aware of this and intentional about where you've placed that code?
- consts behavior has been streamlined so that there is less overlap between theme and finalize, no 4-item padding vectors, and fewer consts overall. 
- `geom_text_lastonly` has new documentation that describes additional solutions to the clipping of labels on the right edge of the plot's panel.
- `theme_cmap_base` is now a base complete theme with only a few arguments and functional inheritance. It is not exported but serves as the basis of `theme_cmap`.
- `theme_cmap` completely rewritten to build a list of gg objects (`obj`) and a list of theme attributes (`attr`) based on user spec. At the end of the function, the objects are returned, as well as an additional `ggplot2::theme()` object with all specified attributes included.
- documentation updated.

Identified to do:
- [ ] update the ppt margins visualization
- [ ] remove in-line definitional commenting on `cmapplot_globals$consts. 
- [ ] confirm non-exported objects are documented correctly (e.g. `theme_cmap_base`)

nmpeterson:
>     * fonts should load upon actual loading of package via `library(cmapplot)`. @nmpeterson, `check()` warns about preferring messages in `.onAttach` rather than `.onLoad`-- just checking that you're aware of this and intentional about where you've placed that code?

From [**R packages** chapter 13.2](https://r-pkgs.org/namespace.html#search-path):

> - **Loading** will load code, data and any DLLs; register S3 and S4 methods; and run the `.onLoad()` function. After loading, the package is available in memory, but because it’s not in the search path, you won’t be able to access its components without using `::`. Confusingly, `::` will also load a package automatically if it isn’t already loaded. It’s rare to load a package explicitly, but you can do so with `requireNamespace()` or `loadNamespace()`.
> 
> - **Attaching** puts the package in the search path. You can’t attach a package without first loading it, so both `library()` or `require()` load then attach the package. You can see the currently attached packages with `search()`.

It's not hard to imagine a case where someone forgets to load the package initially and then decides to add `+ cmapplot::theme_cmap()` to their plot instead of running `library(cmapplot)`. In this case, if the font handling was done by `.onAttach()`, I don't think it would ever be run and the fonts in the plot would all be Arial.

nmpeterson (approved):

dlcomeaux:
@matthewstern FYI, the exported fonts are still "Cairo Font" for anything exported in Calibri Light. However, it appears to be the correct face, so I would say we leave it for now.

dlcomeaux (approved):
This looks fantastic, @matthewstern ! Great work. I've done some testing in an installed version and everything appears to be in working order, with the exception of on-plot labels, which are now inheriting the right font size (yay!) but are still not inheriting the right font face (boo...but I think we may have to address that through an updated version of #67 as mentioned there). 

Also, calling colors like `blackish` now works, as expected.

PR #73, opened 2020-10-30
Font detection in .onLoad()

nmpeterson:
Ok, it took way longer than I expected, but I _think_ I've got the font issues sorted out now. I'd appreciate confirmation from Windows users, though. You can install and load this branch in R as follows:

```
devtools::install_github("CMAP-REPOS/cmapplot", ref="fontsizes-onload")
library(cmapplot)
```

Basically, the Whitney identification procedures got moved into `.onLoad()` (which I kept in cmapplot.R instead of creating zzz.R). In order to make this work, I also had to convert cmapplot_globals from an environment into a list, that is now exported as part of the package. This may help with @dlcomeaux's question in #67, too.


nmpeterson:
Also note that this PR is merging into @matthewstern's "fontsizes" branch and not "master"

R/cmapplot.R
```diff
@@ -17,73 +17,125 @@
...
+      )
+    }
+
+    # If non-Windows machine, stick to Arial
   } else {
-    message("WARNING: Whitney is not installed on this PC, so CMAP theme will default to Calibri")
-    # Add fonts to R
-    grDevices::windowsFonts(
-      `Calibri` = grDevices::windowsFont("Calibri"),
-      `Calibri Light` = grDevices::windowsFont("Calibri Light")
```
matthewstern, line 63: Do Calibri and Calibri light exist on every windows machine? Just wondering if it would be helpful to verify, now that Arial is the package's universal default.

R/cmapplot.R
```diff
@@ -17,73 +17,125 @@
...
+    big = 17,
+    reg = 14,
+    sml = 11
+  ),
+
+  ## Base typefaces -- modified later by .onLoad()
+  font = list(
+    strong = list(family="Arial", face="bold"),
+    regular = list(family="Arial", face="plain"),
+    light = list(family="Arial", face="plain")
```
matthewstern, line 44: I'm still at 50% comprehension with fonts in R, but it's interesting to me that this works considering that WindowFonts() doesn't display a font called "Arial", but rather:

```
$sans
[1] "TT Arial"
```

Is it just different on Macs?

matthewstern (approved):
This looks good and works for me.

If we move forward, there is documentation in `theme_cmap()` and `finalize_plot()` that could move to the new help file for `cmapplot_globals`.


```
> library(cmapplot)
Loading required package: ggplot2
WARNING: Whitney is not installed on this PC, so CMAP theme will default to Calibri
> windowsFonts()
$serif
[1] "TT Times New Roman"

$sans
[1] "TT Arial"

$mono
[1] "TT Courier New"

$Calibri
[1] "Calibri"

$`Calibri Light`
[1] "Calibri Light"
```

```
> econ_plot <- ggplot(data = cluster_jobchange, 
+                     mapping = aes( 
+                         x = reorder(name, jobchange), 
+                         y = jobchange, 
+                         fill = category)) + 
+     geom_col() + 
+     coord_flip() + 
+     theme_cmap(gridlines = "v", hline = 0) + 
+     scale_y_continuous(labels = scales::comma) 
'Whitney' font family not found. Using a substitute... 

> econ_plot + cmapplot:::theme_cmap_base()
```

image

R/cmapplot.R
nmpeterson: [Yes, back through Windows Vista.](https://docs.microsoft.com/en-us/typography/font-list/calibri#products-that-supply-this-font)

R/cmapplot.R
nmpeterson: I'm not sure if it "works" or if it's just silently failing and and then reverting to R's default font... Arial!

dlcomeaux:
I'm getting an error when I attempt to use `finalize` installed from this branch:

```
Error in check.length(gparname) : 
  'gpar' element 'fontsize' must not be length 0
```
Neither of you had anything like that, I assume? @nmpeterson @matthewstern

dlcomeaux:
It looks like this branch has not yet updated `finalize` to incorporate the new font size specifications? I think that's the source of the error. In which case, how did it work successfully for both of you? Sorry if I'm missing something obvious.

nmpeterson:
@dlcomeaux yes, you're right, @matthewstern has not yet updated `finalize_plot()` to work with the revised font specification. In our tests, we were only using `theme_cmap()` (and the new `theme_cmap_base()`, not `finalize_plot`. I just wanted to verify that you no longer get the font errors on Windows when loading the package via `library()`, and that `theme_cmap()` correctly uses Calibri instead of Arial.

dlcomeaux:
OK. Yes, when I use `theme_cmap_base()`, it uses Calibri with no errors!

R/cmapplot.R
```diff
@@ -17,73 +17,125 @@
...
+    margin_legend_b = 10, # [t,f] Margin between legend and plot (within plotbox)
+    margin_plot_b = 5,    # [ ,f] Margin between plotbox and bottom edge of image
+    margin_title_l = 2,   # [ ,f] Margin between left edge of image and title/caption
+    margin_title_r = 10,  # [ ,f] Margin between title/caption and legend/plot
+                          #        (Both title_l and title_r are deducted from `titlewidth`)
+    margin_plot_r = 10,   # [ ,f] Margin between plotbox and right edge of image
+    padding_plot = c(0,10,0,0),
+                          # [t,f] Padding around the entire plot
+    padding_legend = c(0,0,0,0),
+                          # [t,f] Padding for the legend element. The -9.5
```
dlcomeaux, line 75: Flagging to update this documentation (about the -9.5) given the new adjustment making it irrelevant.

R/cmapplot.R
```diff
@@ -17,73 +17,125 @@
...
+                          #        adjustment is necessary to left align legend
+                          #        key with plot elements. Additional adjustments
+                          #        may be required based on system configuration,
+                          #        which can be done easily using `legend_bump`
+    legend_key_size = 14, # [t,f] Size of the legend key item.
```
dlcomeaux, line 80: It may make sense to update this to a variable, instead of hardcoding, to be consistent with the default size for the legend text - i.e, `cmapplot_globals$fsize$reg`

dlcomeaux (approved):
Left a couple flags for minor things to update, but they are tangential to this PR and can be addressed later. Looks good to me and works as expected (once I understood what it was I was testing!).

R/cmapplot.R
matthewstern: Can do. Good idea.

R/cmapplot.R
dlcomeaux: I addressed this in the base branch.

R/cmapplot.R
dlcomeaux: Given that we are going to shift to `$s|$m|$l` this can go on the list of changes we should make after that is implemented.

PR #80, opened 2020-11-12
new method for geomdefaults

matthewstern:
This PR is set to merge into #67. The intention is to isolate the setting and resetting of geom defaults into their own functions that can be called not only in `finalize_plot` but also by the user as to have our custom geom defaults impact ggplots pre-the finalize stage.

Two new exported functions are created, currently in cmap_ggplot_defaults.R: 

- `fetch_cmap_gg_defaults` returns a named list of all the geom defaults we would seek to overwrite (for replacement later)
- `set_cmap_gg_defaults` sets geom defaults for all the geoms we plan to overwrite. This function will either overwrite with our defaults, or will take the list outputted by `fetch` to reset.

The general workflow (which is now implemented in `finalize_plot` but can also be implemented by the user) is:
```
g <- some ggplot

g   # display g with standard settings

old <- fetch_cmap_gg_defaults()
set_cmap_gg_defaults()

g   # display g with our preferred settings

set_cmap_gg_defaults(old)

g   # display g with standard settings
```

To do this, I have created a tibble in `cmapplot_constants` that contains the geoms to overwrite and their preferred values. Here's where things get tricky. Because the preferred values are calculated *based on other other cmapplot values and functions* (e.g. `ggplot_size_conversion` and values from `cmapplot_globals` such as font families and faces) I needed to add this via `.onLoad`. This works pretty well, but it's not perfect. 

**It seems to take the size attributes OK, but it leaves fonts as Arial.** In other words, even though fonts are set before the tibble is created, the values taken from `cmapplot_globals` are still the old ones, maybe because the version of `cmapplot_globals` available within the `.onLoad` environment is still the old one with all fonts set to Arial? 

This is where I could use some help, probably from @nmpeterson. Am I thinking about this the right way? How would I coax `.onLoad` to fetch the updated fonts? Is there somewhere besides for `.onLoad` I should put this?

Here's the test code that I've been using:
```
  geom_line() +
  geom_text() +
  geom_text_lastonly() +
  theme_cmap() +
  coord_cartesian(clip = "off")

# without special defaults
g

# with special defaults
finalize_plot(g)

# without special defaults
g

# with special defaults
old <- fetch_cmap_gg_defaults()
set_cmap_gg_defaults()
g

# without special defaults
set_cmap_gg_defaults(old)
g
```

If we can get this to work, I still need to:

- [ ] write documentation for the new functions
- [ ] add an arg to `finalize_plot` to ignore the set and reset of geom defaults if the user wants
- [ ] although we have long thought that updating geom_defaults would not apply to `finalize_plot`'s return of an object (and our documentation says this is the case), confirm that this is actually wrong because what finalize returns is a grid object, not a ggobject. If correct, move the geom reset higher in the code and edit the documentation.


matthewstern:
Sorry, meant for this to merge into #67.

PR #84, opened 2020-11-18
Documentation tweaks

matthewstern:
a variety of small documentation changes, the most substantial of which is combining help files for `cmap_gradients` and `viz_gradient()`, and same for palettes.


pkgdown/_pkgdown.yml
```diff
@@ -22,17 +22,15 @@ reference:
...
 - contents:
   - starts_with("geom_")
-  - customproto
   - cmap_default_aes
-  - gg_lwd_convert
   - integer_breaks
+  - gg_lwd_convert
   - update_recessions
   - cmapplot_globals
+  - customproto
```
nmpeterson, line 30: I can't figure out why this shows up as "customproto" on the pkgdown reference page (instead of "GeomTextLast", "GeomPointLast", etc.), while "cmap_default_aes" shows up as "apply_cmap_default_aes" & "unapply_cmap_default_aes". It's not a big deal, but it's bothering me a little bit.

matthewstern: Interesting. I wonder if it has to do with none of the objects within customproto being functions? 

The only difference I see in the .Rd files themselves is the `\docType` line. This could drive the change.

```
\docType{data}
\name{customproto}
\alias{customproto}
\alias{GeomRecessions}
\alias{GeomRecessionsText}
\alias{GeomTextLast}
\alias{GeomPointLast}
\title{Custom ggproto classes}
```

vs:
```
\name{cmap_default_aes}
\alias{cmap_default_aes}
\alias{apply_cmap_default_aes}
\alias{unapply_cmap_default_aes}
\title{Fetch and set aesthetic defaults}
```

However, the man file for `.lwd` and `gg_lwd_convert` is listed as a data type, and still shows up as a list on the pkgdown site.

```
\docType{data}
\name{.lwd}
\alias{.lwd}
\alias{gg_lwd_convert}
\title{Line width conversion}
```

In other words, <<shrug>>

Issue #78, opened 2020-11-09
Example charts for cookbook


dlcomeaux:
For the cookbook's fourth vignette, we decided to make a list of example "fully baked" CMAP charts. These should hit each of the major categories of charts that are regularly produced in CMAP publications and analyses. @gritzenthaler @nmpeterson @matthewstern @sarahcmap - please feel free to claim one (or more) of the following, or adding your own ideas, by editing this overview. 

We can also have two examples for some of these, if there are good ones - in particular, given how common line and bar charts are, those might be good to have some extra.

**Sample charts**
- [x] Line chart (a fully built-out version of the one that Matt is developing in his vignette is a good candidate for this)
- [x] A scatterplot (like [this](https://www.cmap.illinois.gov/documents/10180/1034772/PU20-0001_AGE_LABOR_FORCE_Change_in_size+copy.png/5c50d430-a633-97a8-26d9-7ffadadca080?t=1564670176836) one). Ideally, this will demonstrate changing both size and color of the points, as the linked one does.
- [x] An area chart. My current idea would be to do a variant of the Line Chart 1 above and use the transit data in a different way to do this.
- [x] Clustered bar chart: The replicated Census chart that I made should do the trick for this, as it demonstrates adjustments to bar sizes, annotations with text, dashed lines, 
- [x] Stacked bar chart (either absolute or scaled to share out of 100%)
- [x] Some kind of chart using the highlight function (and probably a bar chart). There is a census piece chart that could be modified to do this, if no one has other ideas.
- [x] A histogram
- [x] A pie chart (for example, like [these](https://www.cmap.illinois.gov/updates/all/-/asset_publisher/UIMfSLnFfMB6/content/cmap-estimates-show-improved-regional-air-quality-due-to-recent-transportation-shifts))
- [x] A faceted chart (one of the above could also accomplish this).
- [x] A chart using the race/ethnicity palette. One of the inclusive growth indicators would be a good candidate, e.g. [this](https://cmap-repos.github.io/ONTO2050-indicators/#household-income-race-ethnicity).

matthewstern:
Thanks @dlcomeaux. In PR #79, which should hopefully be merged before we meet again in two weeks with our example code, I've established a cookbook vignette that can serve as the eventual home for these.

nmpeterson:
I took a stab at creating the donut chart on the page @dlcomeaux linked. (Can be turned into a normal pie chart by dropping the `ylim()` call.)

[...]

PR #85, opened 2020-11-19
New vignettes: cookbook, finalize, colors

nmpeterson:
Here is a preliminary version of the cookbook, where I've compiled all of the examples to date in #78. I've tried to keep the formatting (both of the layout and the code itself) pretty consistent across examples, and I have embedded the original published graphics in the cases where we have one.

A few chart types are still missing:

- [x] Line chart. (@matthewstern can you adapt your transit ridership example from plots.Rmd?)
- [x] Histogram. (@gritzenthaler claimed this one...)
- [x] Faceted chart. (... and this one)

There are already sections in cookbook.Rmd where these can be inserted once completed.

<strike>Note that I have been calling `finalize_plot` in PNG mode and then including an HTML `` tag to explicitly embed the output. Calling `finalize_plot` in plot mode works, but the pkgdown layout is too narrow to accommodate the default image size, and using small enough `width` and `height` parameters wreak havoc on some of the plot layouts. The way I have done it, the image is rendered correctly and then scaled down to fit within the pkgdown layout.</strike> 

**Edit:** Matt figured out a good way to allow `mode="plot"` to work correctly, so this is the preferred method and I have converted all existing plots to use it (except one plot in the "finalize" vignette that demonstrates the `mode`/`filename` parameters).

`````markdown
## CHART TYPE

INTRO TEXT

#### Example code

```{r CHARTTYPE, message=FALSE}
df <- #CODE TO CREATE/LOAD/FORMAT DATA

p <- #CODE TO CREATE PLOT

finalize_plot(
  p,
  title = "CHART TITLE",
  caption = "CHART CAPTION"
)
```

#### Original graphic

<img src="URL OF PUBLISHED GRAPHIC" width="100%" style="display: block; margin: auto;"/>

[Source](URL OF WEB PAGE CONTAINING PUBLISHED GRAPHIC)
`````

**EDIT:** As you edit this, you can preview the pkgdown site locally by running `pkgdown::build_site()` after saving your changes to cookbook.Rmd. This takes a minute or two to run, but when it's finished it will open your web browser with a temp version of the site.

### CODE REVIEW

File: vignettes/cookbook.Rmd [hunk 1]
```diff
@@ -21,82 +21,568 @@ library(tidyverse)
...
+#### Original graphic
+
+<img src="https://www.cmap.illinois.gov/documents/10180/854316/MedianIncome_3.svg" width="100%" style="display: block; margin: auto;"/>
+
+[Source](https://www.cmap.illinois.gov/programs/regional-economic-indicators/trends)
+
+
+## Clustered bar chart
+
+Intro text
```
nmpeterson, line 270: @dlcomeaux any intro text you want to add for the clustered bar chart?

README.md
```diff
@@ -88,3 +87,26 @@ finalize_plot(title = "Regional population and labor force participation",
...
+    scales](https://r-graphics.org/recipe-axes-range), [coordinate
+    systems](https://r-graphics.org/recipe-axes-polar), and
+    [facets](https://r-graphics.org/recipe-facet-basic).
+
+  - The [ggplot2 book](https://ggplot2-book.org/) delves deeper into why
+    and how ggplot2 works the way it does, also with distinct chapters
+    on topics like
+    [scales](https://ggplot2-book.org/scales-guides.html), [coordinate
+    systems](https://ggplot2-book.org/coord.html),
+    [facets](https://ggplot2-book.org/facet.html), etc.
```
nmpeterson, line 112: Consider also adding links to [R for Data Science](https://r4ds.had.co.nz) and/or the official [ggplot2 package website](https://ggplot2.tidyverse.org)?

vignettes/finalize.Rmd
```diff
@@ -0,0 +1,162 @@
...
+The function has a number of customization options built in, accepting 16 (or more) arguments. Refer to the object documentation, `?finalize_plot`, for detailed information on all arguments.
+
+In this vignette we will use the final version of the line chart developed in `vignette("plots")`. That plot is referenced in the following examples as `p`.
+
+## Basic implementation
+
+After creating a plot and applying `theme_cmap()`, use `finalize_plot()` to complete the implementation of
+CMAP design standards. You will probably want to set at least the `title` and `caption`, although the function will extract them from the ggplot if they were specified. As you are preparing the plot, you will likely want to view it within R. Do this by leaving leaving the default `mode = "plot"` to send the finished plot to the "Plots" tab within RStudio. Windows users may prefer `mode = "window"`, which draws the plot in a popup window. In both modes, the plot will show up "actual size" (depending on your screen's resolution) surrounded by a grey canvas.
+
+```{r finalize1, fig.width= 11, out.width = "100%"}
```
matthewstern, line 57: @nmpeterson I found a way to scale outputs of `finalize_plot` without relying on png export and re-placement. I've used it here to demonstrate the canvas drawn around the plot in R. I wanted to flag it for you in case you think it's worth implementing elsewhere? I believe this can be set globally at the top of the file.

matthewstern (approved):
When I test the pkgdown site with `pkgdown::build_site()`, the finalize_plot pngs do not show up for me--they work when knitting within RStudio, however, and I am hoping this is just a site effect of how the pkgdown package builds the temporary local site and is not representative of some larger issue I introduced?

vignettes/finalize.Rmd
nmpeterson: This is a great adjustment. I've implemented everywhere except the one plot in `finalize` that is explicitly an example about using the `mode` and `filename` parameters.

vignettes/cookbook.Rmd
```diff
@@ -36,68 +37,72 @@ Line graphs are useful for showing change over time, and are well-suited for com
 
 #### Example code
 
-```{r line, message=FALSE}
-# get example dataset, and create a relative GRP field
+```{r line, fig.asp=4.5/(670/72), message=FALSE}
```
nmpeterson, line 40: @matthewstern FYI, this had to be set in order to use `mode="plot"` and `height=4.5` in `finalize_plot()` without grey bars on the top and bottom of the image.

vignettes/cookbook.Rmd
```diff
@@ -11,10 +11,11 @@ vignette: >
 knitr::opts_chunk$set(
   collapse = TRUE,
   comment = "#>",
-  fig.width = 7,
+  fig.width = 670/72,
```
nmpeterson, line 14: @matthewstern I'm keeping the width at 670/72 for the "finalize" and "cookbook" vignettes to enforce correct layouts. Smaller widths can cause havoc with long titles/captions, especially in cookbook.

vignettes/cookbook.Rmd
```diff
@@ -36,68 +37,72 @@ Line graphs are useful for showing change over time, and are well-suited for com
...
-  # modify default sizes a touch
+  # Tweak plot layout
   height = 4.5,
   title_width = 2.6,
-  
-  # use some <br> tags to get the title and caption to match perfectly
-  title = "Cumulative growth<br>in real gross regional product, 2001-19",
-  caption = "Note: Indext year 2001. <br><br> Source: Chicago Metropolitan
-    Agency for Planning analysis of U.S. Bureau <br>of Economic Analysis and 
-    Economic Modeling Specialists International <br>data (Emsi 2020.4).",
```
nmpeterson: @matthewstern FYI I removed the mid-sentence `<br>`s. While I appreciate that they were adding line breaks in the original locations, it doesn't quite work as intended: 
Screen Shot 2020-11-22 at 2 01 39 PM

I don't think we should be encouraging people to use `<br>` tags unless truly necessary, anyway (e.g. between note and caption). It just feels like we're trying too hard to exactly replicate the original.

vignettes/cookbook.Rmd
```diff
@@ -341,7 +345,8 @@ df <- tibble(
 p <- ggplot(data = df,
             mapping = aes(
               x = reorder(RACE, -RESPONSE), # Order race by decreasing response rate
-              y = RESPONSE, fill = YEAR, label = LABEL
+              y = RESPONSE, fill = YEAR, 
+              label = LABEL # note that this label definition is implicitly referenced by `geom_text()` (below)
```
nmpeterson, line 349: Maybe this should be moved into `geom_text`'s `aes`, since that's the only geom that uses it?

vignettes/cookbook.Rmd
```diff
@@ -11,92 +11,630 @@ vignette: >
...
-```{r bars, eval=TRUE}
-# A bar chart
-ggplot(cluster_jobchange, aes(x = reorder(name, jobchange), y = jobchange, fill = category)) +
-  geom_col() +
-  coord_flip() +
-  theme_cmap()
+
+## Area plot
+
+Area plots are useful for showing changes in composition over time. (See also: stacked bar charts.)
```
sarahcmap, line 110: I see we can link to the stacked bar charts section directly if we want by putting the section name in brackets:

```suggestion
Area plots are useful for showing changes in composition over time. (See also: [Stacked bar chart].)
```

sarahcmap (approved):

vignettes/cookbook.Rmd
```diff
@@ -11,92 +11,630 @@ vignette: >
...
+)
 ```
 
+#### Original graphic
 
-## Sample plots
+<img src="https://www.dropbox.com/s/xol70o61citlpjj/Cumulative_Growth-01.png?raw=1" width="100%" style="display: block; margin: auto;"/>
 
-### Bar chart
+<!--[Source]()-->
```
nmpeterson, line 105: @matthewstern is there a CMAP webpage or PDF that contains this graphic that we can link as the source?

nmpeterson: I've found several variations of the chart on the website, but not the exact one that you've saved on Dropbox.

matthewstern: I'm working on this, I got it directly from Austen so dropbox was my cheap over-the-weekend out. I don't think it's published directly on the website anywhere but I'm hoping we can at least link to a published committee presentation.

matthewstern: Just pushed an imperfect fix for this. That link has the most recent CMAP website version at the bottom.

vignettes/cookbook.Rmd
dlcomeaux: Makes sense, I can adjust.

dlcomeaux (approved):

gritzenthaler:
Sorry for the delay--the facet parts of the graph added more confusion than I thought they would. I've added a faceted histogram, with one issue I can't figure out, and one thing which could or could not be an issue.

- I want to scooch the row facet labels (work and school) a little down because now they're awkwardly high up

- I found one example of a facet graph in our resources (https://www.cmap.illinois.gov/programs/regional-economic-indicators/clusters#Skills_Gap_2017)  that put lines between the row facets, so I'd imagine that's something we'd like to do? I'm stumped as to how to add that as well.

If anyone has some suggestions, I'm all ears.

nmpeterson:
@gritzenthaler I made some adjustments to your code. Setting `hjust=0.5` for the theme's `strip.text` parameter (which controls facet labels) centered the tour purpose labels vertically. I added `hline=0` for visual emphasis of the bottom of each chart.

I also decided to remove the column faceting (by departure vs. return) and symbolize that by fill color instead, since there was so little overlap between departure and return distributions for work and school trips. I added a third row showing discretionary trips for a little extra variety.

vignettes/cookbook.Rmd
```diff
@@ -476,52 +476,74 @@ finalize_plot(
...
+  
+  # Facet horizontally by tour purpose
+  facet_grid(rows = vars(purpose)) +
+  
+  # Apply CMAP theme and color palette
+  theme_cmap(
+    hline = 0,  # Add origin line at 0 for aesthetic reasons
+    gridlines = "hv",  # Show horizontal and vertical gridlines
+    strip.text = element_text(
+      hjust = 0.5  # Center facet labels at axis midpoint
```
matthewstern, line 524: If there are lessons here such as the `hjust` change (or from @dlcomeaux's draft work with transit investments) about how to make `theme_cmap()` handle facets better than they do by default, I don't see an issue incorporating them into `theme_cmap_base()`. 

If we keep the adjustment here, I think it would be helpful to clearly comment that `strip.text` is a `...` arg that is passed on to `ggplot2::theme()`

dlcomeaux:
To @gritzenthaler 's question about lines in between facets - I found [this](https://community.rstudio.com/t/how-do-i-control-facet-borders-with-ggplot2/1178), which gets you somewhat close, but unfortunately doesn't quite work. 

I've played around with it for a while and can't find a way to center the lines between facets, as opposed to resting on the top and bottom of the middle facet.

image

For reference, you get this by adding the following line to the ggplot call:
`geom_hline(data=tibble(purpose = "School", y=c(-1,1) * Inf), aes(yintercept=y), col = "blue") +`

vignettes/cookbook.Rmd
dlcomeaux: My only concern with shifting `hjust` to 0.5 in this case would be that, for labels on top of the graph (say it's a differently oriented facet), a left-aligned label is more consistent with the title and caption's labeling style.

dlcomeaux:
Also - @nmpeterson and @gritzenthaler this is a great looking faceted chart and histogram! Nicely done.

vignettes/cookbook.Rmd
nmpeterson: Yeah, I think this might be good to incorporate into `theme_cmap()`, but I would want a recommendation from Comms about preferred label placement first.

gritzenthaler:
Thank you! @nmpeterson this looks great! I figured the label adjustment would be simple, but I couldn't find the right parameter to reference. 

However, I'm checking the code from your latest commit now, and it doesn't look like it's doing what that picture you just posted is doing. Is there a reason the facets and fill got switched around?

vignettes/cookbook.Rmd
matthewstern: > 
> 
> My only concern with shifting `hjust` to 0.5 in this case would be that, for labels on top of the graph (say it's a differently oriented facet), a left-aligned label is more consistent with the title and caption's labeling style.

`strip.text` is one of those `theme()` args that has many varieties: `strip.text.x|y` and maybe even `strip.text.x|y.left|right|top|bottom` (one of these is in `theme_cmap_base()`, which means I probably borrowed it from somewhere, but strangely it is not [listed as an arg](https://ggplot2.tidyverse.org/reference/theme.html).

But at least it is possible to have a different default for x and y strips.

nmpeterson:
@gritzenthaler yes, I kept tinkering after that message. I swapped the fill and facet variables because including both arrivals and departures in one chart was not yielding a true histogram. I've also made a change to include ALL modes (albeit everything except work/school are grouped as "other"). 

histogram-1-2
I'm done tinkering now.

matthewstern:
Wow this looks so good @nmpeterson @gritzenthaler !  Congrats! Are we ready to merge this in?

nmpeterson:
Ok, *now* I'm done tinkering, I swear. I added a second dimension to the facet for modeled vs. observed data.

histogram-1-3

gritzenthaler:
I went in to make the description changes, but Noel already took care of it; thank you! I just added the strip.text note and source for the observed data.

PR #86, opened 2020-11-19
Rebuild and improve `finalize_plot()`

### LINKED ISSUE #55: improving iteration functionality in finalize_plot
In iteration through multiple save modes, `finalize_plot` will crash if one save fails. For example, if "test.pdf" already exists and is open by the user, so cannot be written to. Better behavior would be to skip that file if opening the device fails. I tried to implement `tryCatch` but couldn't figure it out.

Relatedly, the function currently overwrites past outputs with the same name with no warning or confirmation. Ideally tackled at the same time as the error handling above, it would be nice to introduce a new argument such as `overwrite = TRUE`, and have the function skip any existing file outputs that already exist if the argument is FALSE (the default).

**Edit:** the solution probably looks like this:
```
if(file.exists(file) & !overwrite){
    # warn and skip
  } else {
    tryCatch{
      # attempt the save
      # warn and move on if erroer
    }
  }
```

matthewstern:
Ever since @dlcomeaux implemented the `buildChart` subfn in `finalize_plot`, I've wanted to continue down that path and redesign `finalize_plot` as four distinct “movements”, three of which can be captured primarily in subfns. Addressing #55 gave me an excuse to do this. 

The benefits of this new structure are manifold, I think: the function is more legible/less overwhelming; code redundancy is reduced, especially in the outputting section; it was simpler to address #55; and future implementation of new features (such as a map template version of `finalize_plot`) is hopefully eased.

The diff view of `finalize_plot.R` is hard to interpret. Here's a summary of changes:

**Movement 1: initialization**
- changes elsewhere allowed `filename` to default to `NULL` rather than `""`
- `mode` arg checking reorganized so `mode = NULL` defaults to `mode = "plot"`
- `"window"` mode removed from list of modes on on-Windows machines (hoping @nmpeterson can test)

**Movement 2: modifying the ggplot**
- `buildChart` function renamed now `prepare_chart`, now includes theme modification work that had been done outside of function before
- new `tryCatch` implementation allows default_aes to be reset even in the case of error in `prepare_chart`

**Movement 3: constructing the final layout**
- new `construct_layout` subfn isolates the construction of the final plot. This subfn now creates the `plotbox` viewport and all the necessary in-plot grobs, and then stitches them together.

**Movement 4: outputting**
- drawing in-R handled first via iteration. Now handled in subfn `draw_plot`, which compresses `plot` and `window` mode into a single implementation. This subfn draws the canvas grob, implements and pushes the centerframe viewport, and draws the plot.
- exporting now handled in subfn `save_plot`, which is run first for vectors and second for rasters. This function has a few new features:
  - checks if file exists and overwrites only if new arg `overwrite = TRUE`
  - `tryCatch` safely moves on if file write fails
  - new, consistent confirmation messaging (closes #55)
- the plot is now always returned as a gTree object. `mode = "object"` is no longer an option. This happens implicitly (using `invisible()` rather than `return()`, so won't bother the user.


R/finalize_plot.R
```diff
@@ -135,21 +140,25 @@ finalize_plot <- function(plot = NULL,
...
                                 savetypes_vector,
-                                savetypes_print,
-                                "object"),
+                                savetypes_raster),
                     several.ok = TRUE)
 
+  # remove any `window` mode specified if OS is not Windows
+  if (.Platform$OS.type != "windows"){
+    mode <- mode[mode != "window"]
+  }
```
nmpeterson, line 157: I'd like to request different behavior here, if it's not too much work. It's not great practice to implement arguments in a way that code will work on one system and then cause a hard crash if run on a different one. Here's my suggestion: allow `mode="window"` as a valid parameter on non-Windows, but switch the mode to `"plot"` at runtime and produce a warning like "mode='window' is not supported on non-Windows systems. Switching to mode='plot' instead."

matthewstern: Does this cause a crash? The goal was to simply remove "window" from the list of modes and continue with the function otherwise. This would mean no modes are listed, I guess, if  "window" is the only argument passed. Maybe that's where the crash comes from? If so, the easiest solution may be to move this conditional to above `match.arg` and add the warning. That way, `mode = NULL` will be converted to `mode = "plot"` by `match.arg` if it is not already in the list.

Happy to tweak... can you clarify what behavior you're currently experiencing though?

nmpeterson: I hadn't tested and was assuming it would fail if I ran `finalize_plot(mode="window")` on macOS. I just tested and it didn't crash *per se*, but no plot was generated and no message was provided. Here's my revised request:

- If `mode` includes `"window"` but not `"plot"`, replace `"window"` with `"plot"` so that a plot is generated, and provide a warning.
- If `mode` includes `"window"` *and* `"plot"`, remove `"window"` from the list (current behavior). Also provide a (different) warning in this case.

matthewstern: Let me know if the commits I just pushed work well on your system

nmpeterson: Works as advertised

nmpeterson (approved):

dlcomeaux (approved):
Looks great. Love the handling of `overwrite` and PDFs.

PR #100, opened 2020-12-04
Tick marks

dlcomeaux:
This PR adds an argument to `theme_cmap()` that allows for the easier addition of tick marks on both the x and y axes. It includes a note in the documentation about the edits required for them to touch axes (involving a manual edit of `expand` inside the appropriate `scale` call). I didn't think it made sense to have that adjustment to `expand` automatically happen, since it is unrelated to tick marks and there are situations where you might not want it, but open to other thoughts there (and if so, ideas on how to implement it).

If approved, we'll want to modify the pkgdown section on this accordingly.

### DIFF
--- a/R/cmapplot.R
+++ b/R/cmapplot.R
@@ -101,7 +101,7 @@ cmapplot_globals <- list(
     lwd_strongline = 1,
     lwd_plotline = 3,
     lwd_topline = 2,
-    length_ticks = 10,
+    length_ticks = 7,
     margin_topline_t = 5,
     margin_title_t = 5,
     margin_title_b = 5,
--- a/R/cmapplot.R
+++ b/R/cmapplot.R
@@ -43,6 +43,7 @@
 #'    \item \code{lwd_topline}: The width of the line above the plot and title. \strong{(F)}
 #'    \item \code{margin_topline_t}: The margin between the top edge of the
 #'    image and the top line. \strong{(F)}
+#'    \item \code{length_ticks}: The length of the axis ticks (if shown).\strong{(T)}
 #'    \item \code{margin_title_t}: The margin between the top line and the
 #'    title. \strong{(F)}
 #'    \item \code{margin_title_b}: The margin between the title and the caption. \strong{(F)}
@@ -100,6 +101,7 @@ cmapplot_globals <- list(
     lwd_strongline = 1,
     lwd_plotline = 3,
     lwd_topline = 2,
+    length_ticks = 10,
     margin_topline_t = 5,
     margin_title_t = 5,
     margin_title_b = 5,
--- a/R/theme_cmap.R
+++ b/R/theme_cmap.R
@@ -114,15 +114,11 @@ theme_cmap <- function(
   obj <- list()
   attr <- list()
 
-
   # create a helper function to more easily add items to the obj list
   add_to_obj <- function(newitem){
     obj <<- append(get("obj", parent.frame()), list(newitem))
     NULL
   }
-
-  # add base theme to object list
-  add_to_obj(theme_cmap_base(consts = consts, debug = debug))
 
   # introduce x label, if specified
   if(!is.null(xlab)){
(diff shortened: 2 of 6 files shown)


matthewstern:
@dlcomeaux would you be willing to throw a few example plots using the feature into this thread?

dlcomeaux:
Sure. The easiest example is the one that Noel references [here](https://cmap-repos.github.io/cmapplot/articles/plots.html#further-theme-modifications), which we can now do more simply:
```

df <- transit_ridership %>% 
  filter(system != "pace_ada") %>% 
  mutate(system = recode_factor(system,
                                cta_bus = "CTA bus",
                                cta_rail = "CTA rail",
                                metra = "Metra",
                                pace = "Pace"
  ))

p <- ggplot(data = df, 
           mapping = aes(x = year, y = ridership, color = system)) +
  scale_y_continuous(breaks = c(0, 200, 400), 
                       limits = c(-1, 575),
                       expand = c(0, 0)) + 
  theme_cmap(ylab = "Annual Ridership (Millions)",
             axisticks = "x") +
  geom_line()

finalize_plot(p)
```
That outputs this:
image

dlcomeaux:
And here's one showing how to do the same with the y-axis.

Code:

```
df <- dplyr::filter(traded_emp_by_race, 
                    variable %in% c("SpecializedTraded",
                                    "UnspecializedTraded"))

q <- ggplot(df, aes(y = reorder(Race, -value), x = value, fill = variable)) +
  geom_col(position = position_stack(reverse = TRUE)) +
  scale_x_continuous(labels = scales::percent,
                     expand = c(0,0)) +
  theme_cmap(vline = 0,
             gridlines = "v",
             axisticks = "y")

finalize_plot(q)

```

image

R/theme_cmap.R
```diff
@@ -178,6 +186,26 @@ theme_cmap <- function(
     ))
   }
 
+  # Introduce x axis ticks if specified
+  if (grepl("x", axisticks)) {
+    add_to_obj(ggplot2::theme(
+      axis.ticks.x = ggplot2::element_line(
```
matthewstern, line 192: certainly not your fault presuming I'm bet you copied the formatting from the `axislines` code above, but the intention in this function was to use `add_to_obj()` for non-theme elements, and use the named list function for theme elements (See, e.g. `attr[["axis.title.x"]]` in the `xlab` section). 

While the end result is not different, it keeps the results of `theme_cmap()` clean and easy to understand. The result of this function should be a list where the first object is `theme_cmap_base()`, the next n objects are non-theme objects, and then there's a single theme object at the end with any additional theme customization. (see, e.g., `View(theme_cmap(gridlines = "none"))`)

Come to think of it, this line at the end could be only added if `len(attr)>0` to simplify further:
```
# construct final list to return
append(obj, list(do.call(theme, attr)))
```

Let me know if you want to make these changes, or I can jump in and make them.

matthewstern (commented):
This looks great to me @dlcomeaux, thanks so much for making them! I just had a few small comments.

R/cmapplot.R
```diff
@@ -100,6 +101,7 @@ cmapplot_globals <- list(
     lwd_strongline = 1,
     lwd_plotline = 3,
     lwd_topline = 2,
+    length_ticks = 10,
```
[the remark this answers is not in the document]
dlcomeaux: I used what was in the vignette I referenced. No idea what Comms' preference is (thankfully that is an easy change).

R/theme_cmap.R
dlcomeaux: Ah. Yes, I did copy and paste. I can take a look at updating this Monday.

R/cmapplot.R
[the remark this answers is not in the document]
dlcomeaux: So yes it would be good to add to the list of things for Ben to weigh in on (should we do that piecemeal or in batches?).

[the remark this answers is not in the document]
matthewstern: Are there other things currently on the list?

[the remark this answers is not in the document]
dlcomeaux: Off the top of my head:
- Tick mark sizing (and line width, I used the standard gridline one but that could also be adjusted)
- Color palettes
- Confirming default highlight colors
- Asking about the top line being over the whole chart vs. just over the title

R/theme_cmap.R
dlcomeaux: It turns out that doing this change actually does impact functionality. If we add things as attributes, if the user then adds an argument that overrides one of those elements (say, `panel.grid.major.y` as I did in the vignette) then it gives an error:
```
Error in (function (line, rect, text, title, aspect.ratio, axis.title,  : 
  formal argument "panel.grid.major.y" matched by multiple actual arguments 
```
Still investigating but flagging in case you know of an obvious solution.

matthewstern: Good catch. We'd just need to separate the overrides into their own `theme()` call that comes last.

I guess that means the returned value of `theme_cmap()` should look something like this:

1. n number of non-theme gg objects (hlines, etc)
2. `theme_cmap_base()`
3. `theme(attr)`
4. `theme(...)` where `...` are additional overriding arguments passed to `theme_cmap()`

Right now # 2 happens first, which is not super relevant. The important thing would be the order of the three theme elements.

dlcomeaux: OK I've incorporated this. Please take a look and let me know what you think.

matthewstern (approved):
@dlcomeaux this looks great to me. In my single commit I made a few changes--some simplifications and some comment clarification, mostly. I also changed the default tick length to 7, which is the same value as `half_line`, a commonly used measure in `theme_cmap_base()`. I think this is probably too long still, but it will suffice until we get more detailed comms review. 

For my testing, I've taken the example plot from `?theme_cmap` and removed the theme elements:
```
g <- ggplot(grp_over_time, aes(x = year, y = realgrp, color = cluster)) +
    geom_line() +
    scale_x_continuous(breaks = scales::breaks_pretty(11))
```

Then, I experimented with a wide variety of implementations of `theme_cmap()`, from the most basic thing the function can produce...
```
t <- theme_cmap(gridlines = "none")
```
image

...to something that just about maxes out the sorts of things the function can produce...
```
t <- theme_cmap(xlab = "test",
                ylab = "test2",
                legend.max.columns = 3,
                axisticks = "x",
                axislines = "x",
                hline = 0.1,
                vline = 2015,
                axis.ticks.x = element_line(color = "red"),
                panel.background = element_rect(fill = "blue"))
```
image

I inspected each with `g + t` and `View(t)` and all seems to work well and look right as per [this comment](https://github.com/CMAP-REPOS/cmapplot/pull/100#discussion_r537639961_) above. 

If you are OK with my tweaks (and maybe re-run a test or two from your end), I'm good to merge.

@nmpeterson should we be updating the package version number with every PR at this point?

nmpeterson:
> @nmpeterson should we be updating the package version number with every PR at this point?

We should formalize our rules for this, but I think the short answer is yes. This is essentially what I envision for version numbering:

- **Major release (e.g. v2.0.0):** reserved for when truly major changes are introduced, especially if they cause compatibility issues with older versions (i.e. the same code would produce different results with the two versions, or old code would no longer work). I'm not sure cmapplot will ever get to this point, unless we decide to expand it to include mapping functionality. (Or we could put that in another package called... "cmapmap"!)
- **Minor release (e.g. v1.1.0):** used to indicate significant _new_ functionality. For cmapplot, I think this would be appropriate if we add a set of new geoms, for example. I think this would also be appropriate for changes to the CMAP design guidelines, where we need to overhaul the appearance of our plots.
- **Patch release (e.g. v1.0.1):** bug fixes and minor new functionality. Essentially any new published version of the code that doesn't warrant a minor or major release number. Adding support for tick marks would make sense here.

And as I've mentioned before, these numbers are not decimals. If we hit 1.0.9, the next release should be 1.0.10 unless it is actually a minor release.

dlcomeaux (commented):
All these changes seem reasonable to me. Thanks!

PR #101, opened 2020-12-16
Creation of a vignette for installation on CMAP computers

dlcomeaux:
A new vignette that outlines the steps needed to install cmapplot on a CMAP-issued computer - it isn't terribly complex, but it does require a few installation steps (specifically, getting Rtools).

### DIFF
--- a/README.Rmd
+++ b/README.Rmd
@@ -50,6 +50,8 @@ devtools::install_github("CMAP-REPOS/cmapplot", build_vignettes=TRUE)
 library(cmapplot)
 ```
 
+For more detailed information about installing the package, particularly on a CMAP-issued computer, see [this article](https://cmap-repos.github.io/cmapplot/articles/installation.html).
+
 To install on macOS, users must install [XQuartz](https://www.xquartz.org) before cmapplot can be loaded. (This can be easily accomplished via the [Homebrew](https://brew.sh) package manager with the command `brew cask install xquartz`.)
 
 **A note about fonts**: The cmapplot package works best when installed on a computer with the Whitney family of fonts installed (specifically the Book, Medium, and Semibold variants). If installed on a computer *without* Whitney, the package will still work, but the fonts will default to Calibri (on Windows) or Arial (on macOS/Linux).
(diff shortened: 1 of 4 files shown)


matthewstern:
Great idea! What do you think about generalizing this away from cmap machines (but keeping notes about what does and does not require admin privileges), and including font installation instructions here as well?

nmpeterson:
I think it's more important to focus on CMAP machines specifically, which will have a consistent and predictable environment and where, importantly, the user will not have administrative privileges. If there's an important note for non-CMAP users, we can put it in the readme (as we did with macOS/XQuartz).

nmpeterson:
When I get my laptop on Monday, I'll go through the vignette and see if I notice anything that's missing.

MMenninger:
I am doing a fresh install on new laptop. I dont even have tidyverse yet. First error calls for RcppEigen. I am going to install tidyverse and start over to see if I get same error.

MMenninger:
After installing tidyverse, I get an error about "pbkrtest" package had non zero exit status.

MMenninger:
I installed "backports" and now everything works.

matthewstern:
Interesting! Glad it's working.

I also set up this morning. I installed rtools first, then installed tidyverse and devtools. Both packages asked to install some packages from source, which worked for me (I presume because I have rtools installed). I was able to install cmapplot without a problem after that.

matthewstern:
Interesting! Glad it's working.

I also set up this morning. I installed rtools first, then installed tidyverse and devtools. Both packages asked to install some packages from source, which worked for me (I presume because I have rtools installed). I was able to install cmapplot without a problem after that.

nmpeterson:
Alright, I worked through this with my new laptop today and, after a few edits, I think it's now good to go.

matthewstern (commented):
The vignette looks great to me. 

Even if we do not generalize this to installing on any CMAP machine, is it reasonable for text on the readme to be coordinated with this, or at least to reflect that this vignette now exists? For example, this vignette says the user needs rtools, while the readme mentions XQuartz but not rtools.  We could also link to this vignette from the readme... something like, "for more information about installing the package, see `vignette("installation")`"?

nmpeterson:
I agree that a link from the readme would be useful, although I would be inclined to link directly to the pkgdown article since, if they have not yet installed cmapplot, they will not be able to run `vignette("installation")`. (I assume your intent here is to take advantage of pkgdown's auto-linking functionality, but it could be confusing in this case because that linking will not work from the repository's main page, where the readme text is also displayed.)

matthewstern:
@nmpeterson yeah, that's right--I was playing off the auto-linking but if you think an html link is better I'm totally fine with that. Come to think of it, I imagine `vingette()` wouldn't work from the github page, even if it would from the pkgdown page.

nmpeterson:
That's correct. The only way to add the link from GitHub would be to use an explicit markdown or HTML link, in which case we may as well avoid the potential confusion and not refer to it as `vignette("installation")`.

nmpeterson:
I've added a link to the vignette's eventual URL (which doesn't exist yet, but will once this PR gets merged and the pkgdown action runs).

matthewstern (approved):

PR #104, opened 2021-01-15
Use Whitney in pkgdown action!

### LINKED ISSUE #92: pkgdown site graphics are Arial-only
For our convenience, I set up a ["GitHub Action"](https://github.com/CMAP-REPOS/cmapplot/actions/runs/373302547/workflow) that rebuilds the pkgdown site every time a commit is pushed to master (including PR merges). This action was developed by the [r-lib team](https://github.com/r-lib/) (who developed pkgdown, devtools, usethis, roxygen2, etc.), and was set up to run the site-building functions in a virtual macOS environment operated by GitHub. (From what I can tell, a new environment is created from scratch each time the action is run, and deleted upon completion.)

The obvious issue with this, for our purposes, is that the graphics generated for our website use the non-Windows font settings (i.e. Arial-only). I think it would be possible to reconfigure the action to use a Windows environment instead (which the [R-CMD-check action](https://github.com/CMAP-REPOS/cmapplot/actions/runs/380023068/workflow) already does), if we want the graphics to use Calibri instead of Arial.

If we want the graphics to use Whitney (once we have access to it), I think the only option will be to disable the pkgdown action and rebuild the site manually from our own PCs using `pkgdown::deploy_to_branch` each time we update master. This seems like a pain in the ass, and is likely to be forgotten for minor commits. Is the trade-off worth it? What are others' thoughts on this?

nmpeterson:
Now that we've changed our GitHub billing plan, I was finally able to grant the pkgdown action access to the Whitney font files (stored in a private repo). I did some extensive testing with the pkgdown-test action (most recent run [here](https://github.com/CMAP-REPOS/cmapplot/runs/1711160140?check_suite_focus=true)), with output saved in the gh-pages-test branch -- both of which should be removed after this PR is merged.

When this PR is merged, the updated pkgdown action will be triggered, and gh-pages will be updated at that time.

Resolves #92


matthewstern (approved):
This all makes sense to me, although the code is beyond me for the most part. I'll merge this and then see what happens with the github pages site.

PR #107, opened 2021-01-28
Allow for charts with no titles or captions

dlcomeaux:
This PR improves the case when `title_width = 0`, removing the otherwise added default text that "This plot needs a title". It also removes any titles or captions that have been manually defined in that instance, warning users that these are not supported with `title_width = 0`.


nmpeterson:
I want to review one thing relating to captions

vignettes/finalize.Rmd
```diff
@@ -143,6 +143,16 @@ finalize_plot(plot = p,
               debug = TRUE)
 ```
 
+### Exporting a graph with no title or caption
+
+One special case of the top-level position adjustments is the elimination of the title and caption altogether. By setting `title_width = 0`, you can export a graph with just the plot and legend elements. Note that in this situation, the constant `margin_title_l` is no longer relevant. `margin_plot_l` is now the margin between the left side of the image and the beginning of the plot (See the next section for more details about these constants).
+
```
nmpeterson, line 149: Do we want to include a caption in this example, to show that it's possible? Or at least mention that captions are still supported with this option?

nmpeterson (commented):
I'm fine with all of these changes, pending a decision about handling captions via #108. If we merge that PR, I think the finalize vignette should be tweaked to either display a caption or mention that they're still supported.

matthewstern (approved):
If it's OK with y'all @nmpeterson @dlcomeaux I propose to merge this PR without #108 or #109 at the moment. 

Noel's work in #108 proves out something we thought might be possible, which is cool. But I don't believe in-plot captioning is strictly necessary for what we'll need to use the package for this week, as captions can be inserted in a Word table row below each graphic. I also don't think it's aligned with the long term direction of the package -- as we think about wanting to expand support for plot objects outside of ggplot2 (e.g. tmap) -- so I don't think we should merge it.

Moving towards a bottom caption grob as Daniel demonstrated in #109 is I think the right direction. But that implementation right now involves quite a lot of repetition, much of which I believe can be avoided. I have some thoughts about how to implement this, which I may be able to work on some this week. But, either way, I think we have what the FFF team needs for now here in this PR.

Are y'all OK with this choice?

nmpeterson (approved):

PR #108, opened 2021-01-29
Use in-plot caption when title_width=0

nmpeterson:
I added this in a separate branch in case we want to go the grob route instead. This works as I intended, and may be a fine short term solution.

HTML formatting does not work with this method, unfortunately.

Screenshot 2021-01-28 185629

### DIFF
--- a/R/finalize_plot.R
+++ b/R/finalize_plot.R
@@ -149,13 +149,6 @@ finalize_plot <- function(plot = NULL,
     message("`mode='window'` is not supported on non-Windows systems. Switching to `mode='plot'` instead.")
   }
 
-  # remove titles and/or captions if title_width is set to 0
-  if (title_width == 0 & (title != "" | caption != "")){
-    title <- ""
-    caption <- ""
-    message("Titles and captions are not supported when title_width = 0.")
-  }
-
   # check mode argument
   savetypes_raster <- c("png","tiff","jpeg","bmp")
   savetypes_vector <- c("svg","ps","pdf")
@@ -204,10 +197,9 @@ finalize_plot <- function(plot = NULL,
     )
   )
 
-  # If title/caption unspecified, try to extract from plot, unless the title
-  # width is set to 0, in which case it will remain blank.
+  # If title/caption unspecified, try to extract from plot
   input_title <- plot$labels$title
-  if (title == "" & title_width > 0) {
+  if (title == "") {
     if (!is.null(input_title)) {
       title <- input_title
     } else {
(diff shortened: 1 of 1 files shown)


R/finalize_plot.R
```diff
@@ -316,22 +325,35 @@ finalize_plot <- function(plot = NULL,
...
+    plot <- plot + ggplot2::theme(
+      plot.title = element_blank(),  # Remove in-plot title
+      plot.caption = element_blank(),  # Remove in-plot caption
+      # apply any extra `ggplot2::theme()` args
+      ...
+    )
+  } else {
+    plot <- plot + ggplot2::theme(
+      plot.title = element_blank(),  # Remove in-plot title
+      plot.caption = element_text(hjust = 0),  # Left-align in-plot caption
```
matthewstern, line 352: This shouldn't be necessary, right? `plot.caption` is already set in `theme_cmap_base()`. If captions should be left-aligned, we should make the change there and allow it to pass through here just by the absence of `plot.caption = element_blank()`. Doing this would allow the user to override the caption justification in `theme_cmap()`, which this implementation does not allow for.

matthewstern (commented):
Hmm... I'm pondering. It adds a decent chunk of complexity that might get in the way if we implement future versions that allow for multiple layout types or multiple input types. e.g. a tmap-based map in `vp.plotbox` might not support this sort of caption modification--and it seems like it would likely be much easier to handle all these layout issues in `grid`. 

All that said, this does appear to solve the problem for right now, which is a boon for the current project. 

What do others think? I'm tempted to go with it, and know we can reverse the change later if we need.

R/finalize_plot.R
nmpeterson: I have no issue with that change

nmpeterson:
I think the question is: do @matthewstern & @dlcomeaux have time to do it "right" (i.e. with a different layout using `grid`) in time to use for the current project? If so, I think we should scrap this PR, given its major drawbacks (no HTML support, and limited flexibility in positioning -- I suspect Comms would want text left-aligned with entire graphic, not just y-axis?).

But, if we need a working solution ASAP to give more time to develop the "right" functionality, then go ahead and merge.

dlcomeaux:
The other option for the current project is for captions to be added below the figures in Word. That's also not ideal, but I think would be preferable in some ways given that it allows for formatting of style, which I'm guessing will be desired.

I will look at the complexity of making a new layout function this morning. I'd vote to hold off on merging this for now.

PR #109, opened 2021-01-29
Vertical layout

dlcomeaux:
This adds the ability to have a vertical layout. It's not done, and in particular, I just got an error message when running a chart with nothing in the `layout_style` argument, which I'll need to look into. But it (mostly) works.


nmpeterson:
So, I assume this should be closed in favor of #111?

dlcomeaux:
Yes, that makes sense to me.

PR #110, opened 2021-02-02
add geom_label to default aes

matthewstern:
This PR adds `geom_label` to the list of geoms that cmapplot overwrites default aesthetics for. Note that I can't get default_aes to modify the special arguments `label.padding`, `label.r`, and `label.size` -- it appears that these are not aesthetics, technically, but arguments of the geom_label function directly. So, we could modify these with custom functions like `geom_label_cmap()` if we wanted, but this seems unnecessary to me. 

This advances version to 1.0.4.

### DIFF
--- a/DESCRIPTION
+++ b/DESCRIPTION
@@ -1,6 +1,6 @@
 Package: cmapplot
 Title: CMAP Themes and Color Palettes
-Version: 1.0.3
+Version: 1.0.4
 Authors@R: c(
     person("Matthew", "Stern",
            role = c("aut", "cre"),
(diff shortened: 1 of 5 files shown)


nmpeterson:
Creating a wrapper function `geom_label_cmap()` could be basically a 1-liner: `geom_label_cmap <- function(...) { ggplot2::geom_label(label.padding = x, label.r = y, label.size = z, ...) }`

matthewstern:
Understood, but the only likely default override is `label.size = 0` to just place the label over a white rectangle -- the defaults for label.padding seem reasonable and the rounded corners from `label.r`'s default are not only relatively unobtrusive but irrelevant when all the user is doing is floating labels over a gridline.

I'd argue for not adding complexity to the package to kist replace the user invoking a single argument to override a default. But, I'm open to other thoughts on that.

nmpeterson (approved):

PR #111, opened 2021-02-02
Vertical layout -- a different approach

matthewstern:
This is inspired by #109 but takes a different approach, in which I adjust the overall flow of the finalize_plot() function to account for the new needs placed on it. My goal is to reduce redundancies to near zero and keep the function highly legible. 

~~This is a work-in-progress. I'm trying to keep the commit descriptions clear, but I'll eventually update this post with a clear description about what this tinkers with.~~

**EDIT:** I think I have succeeded. This PR nets only a few new lines of code (not counting NEWS.md), but both adds features and eliminates redundancy. Detailed changes down below in [this comment](https://github.com/CMAP-REPOS/cmapplot/pull/111#issuecomment-782779446).


matthewstern:
Current progress: No conditionals built in yet, but at the moment the bottom caption grob is in place and the plotbox adjusts accordingly. This seems to work very well where `legend_shift = FALSE`:
image

But something I haven't caught yet is oversizing the grob table where `legend_shift = TRUE`:
image

I haven't fully thought through the conditional yet but I think I'd like to leverage our experience that suggests the title and caption grobs functionally disappear when `title_width = 0`: we don't need to actually remove them. So, I believe that the only conditional we'll need will be adding in the bottom caption when `title_width = 0` and adjusting the plotbox height calculation so it replaces `grobHeight(grob_caption_bottom)` with 0 when that grob does not exist.

matthewstern:
This is ready for review and testing. This makes a variety of under-the-hood improvements to `finalize_plot()` that not only allow for below-plot caption placement but also clean up and streamline the function to pave the way for future implementations, such as tmaps. Here's a summary of changes:

- the `construct_layout()` subfn has been eliminated, which means the creation of grobs now occurs in the main function again. The `prepare_chart()` subfn has been streamlined by incorporating error handling into the function. This is also now called much later in the function, when it's needed, to allow the plotbox height to be calculated later on.  
- plotbox height is now calculated after the caption grob is created. This allows for the plotbox to be drawn smaller as per the space needed for the caption.
- there is a new variable specified in the function (`vert_layout` for now) that specifies whether the title+left caption or the bottom caption should be drawn. *For now, this is determined behind the scenes, when `title_width = 0`. This PR introduces no new user-facing function arguments.*
- The already-existent title and left captions are now built only if `vert_layout = FALSE`, which allowed for the elimination of code that set `title = ""` when `title_width = 0`. The new bottom caption is only built if `vert_layout = TRUE`. 
- grobs are now built into a list. e.g. the object that used to be `grob_title` is now `grobs$title`. This, importantly, allows for an unspecified set of grobs to be easily drawn into the `final_plot` grobTree object. 
- new subfn `safe_grobHeight()` converts grobHeights to bigpoints (or anything else), and returns 0 of that grob doesn't exist, which substantially simplifies dimension specifications that rely on measurements of the heights of other grobs. 

## Sample outputs

(`p` is the code taken from the setup code chunk in the finalize vignette.)
```
finalize_plot(plot = p,
              title = "Annual unlinked passenger trips (in millions).",
              caption = "Source: Chicago Metropolitan Agency for Planning
                analysis of Regional Transportation Authority data.",
              title_width = 1.8,
              caption_valign = "top",
              legend_shift = FALSE,
              debug = TRUE)
```
image

```
finalize_plot(plot = p,
              title = "Annual unlinked passenger trips (in millions).",
              caption = "Source: Chicago Metropolitan Agency for Planning
                analysis of Regional Transportation Authority data.",
              title_width = 0,
              caption_valign = "top",
              legend_shift = FALSE,
              debug = TRUE)
```
image

nmpeterson:
Just a reminder that we'll need to:

- [ ] change the date for release 1.1.0 in NEWS.md whenever this PR gets merged.

nmpeterson (approved):

R/finalize_plot.R
```diff
@@ -216,59 +230,167 @@ finalize_plot <- function(plot = NULL,
...
+                      fontfamily = cmapplot_globals$font$light$family,
+                      fontface = cmapplot_globals$font$light$face,
+                      lineheight = consts$leading_caption,
+                      col = cmapplot_globals$colors$blackish),
+      box_gp = grid::gpar(col = ifelse(debug, "red", NA),
+                          fill = NA)
+    )
+  }
+
+  # in vertical mode, and if caption exists, create bottom caption box (but no title)
```
dlcomeaux: I understand the rationale for this - however, I think that it may be helpful in the future to have the ability to have a title in the vertical layout, potentially above the bar (which was what I had implemented, however jankily, in #109 ). It would create some more complexity, because then you need to have an adjustment to the top line positioning - although I think we could accomplish that with a similar use of the safe_grobheight function you've developed (that one is nice, by the way). Not an urgent function, but something to think about for the future.

dlcomeaux (approved):
I made some minor formatting tweaks and have one suggestion for future functionality (adding titles to vertical layouts) which others may have different ideas on. I think that this is a great upgrade to the function, though - great work, @matthewstern .

dlcomeaux:
@matthewstern @nmpeterson I'm not sure what has happened here - the Windows checks are failing on GitHub, but are not failing when I run them on my computer. The error seems to have something to do with a package called `ggsignif` in the installation (which is failing to install), but that could be incidental. Not sure what's up, since all my changes were in commented text.

dlcomeaux:
Hmm and it's not the result of my commit, since the check still fails when I reverted it. Not sure what's up.

matthewstern:
> 
> 
> @matthewstern @nmpeterson I'm not sure what has happened here - the Windows checks are failing on GitHub, but are not failing when I run them on my computer. The error seems to have something to do with a package called `ggsignif` in the installation (which is failing to install), but that could be incidental. Not sure what's up, since all my changes were in commented text.

@dlcomeaux, thanks for looking! I ran into a similar issue with the auto-checks a few PRs ago. It [looks like](https://cran.r-project.org/web/packages/ggsignif/index.html) the package `ggsignif` was published yesterday to 0.6.1, but that CRAN has not yet updated the package binaries. Last time the error we were getting was that the package was just not available at the version the check VM wanted. This error seems a bit more complex than that, but considering the version difference and recent update, I'm hopeful that's the problem. It seems to take CRAN a few days to do whatever it is they do.

Considering the Mac check passes (and Windows checks pass on both our computers, I think it's reasonable to merge this and re-check the master branch in a few days to verify no issues. Does that seem reasonable? (@nmpeterson)

R/finalize_plot.R
matthewstern: Agreed. With `safe_grobHeight()` and the new conditional grob build framework, it shouldn't be hard to install an alternate title grob above the topline. If you wanted to put together a PR building off this code, I don't see a reason not to implement that too.

nmpeterson:
I'm re-running the jobs right now. If they fail again, I do think it's fine to force the merge. As @matthewstern noted, these issues can crop up occasionally when the virtual machines are setting up their R environment (which is done from scratch each time a job is run) and new versions of dependency packages have recently been added to CRAN but the compiled binaries for those packages are not yet ready.

R/finalize_plot.R
matthewstern: Agreed. My apologies for missing that part of your work in #109. With `safe_grobHeight()` and the new conditional grob build framework, it shouldn't be hard to install an alternate title grob above the topline. It would be really easy, I think... if a new grob top_title (or maybe we change to `title_vert` and `caption_vert`) gets created before the topline does, the topline height could be set to the top margin plus `safe_grobheight(top_title)`...

If you wanted to put together a PR that (re)implements this, I don't see a reason not to implement it.

matthewstern:
FWIW the pkgdown build also failed after merging, due maybe to a similar issue in package "broom"? https://github.com/CMAP-REPOS/cmapplot/runs/1972107044

Hopefully both issues will resolve themselves in the coming days.

nmpeterson:
It's pillar, actually. Note the [CRAN page](https://cran.r-project.org/web/packages/pillar/index.html) currently lists the version as 1.5.0 but the binaries are still 1.4.7

matthewstern:
image

 :-(

Issue #89, opened 2020-11-23
Fix `update_recessions()`


matthewstern:
Looks like NBER has officially placed the US in a current recession (has no close-date), which is throwing off the function. Also, they upgraded their website and changed the URL location for the excel file. Quick fix applied in #85, but should revisit and make sure is working properly. Looks like the open-ended current recession is not processing correctly. 

In addition, tryCatch methods should be implemented within the function so it doesn't hard-error, which will require revisiting the code within `geom_recessions()` (and presumably removing the TryCatch code implemented there)

PR #122, opened 2021-03-17
Fixes bug, allows for current recessions

dlcomeaux:
This PR will address the bug noted in #89 . It updates the placement of error catching in the function to allow for the function to pass `check()` even if the NBER directory changes. However, it also adjusts to use a new NBER URL which should be more stable.

Along with those changes, the PR adds additional functionality:
- Ongoing recessions can now be displayed. Since these have an undefined end date, the new code shifts their end to 2200. Existing code truncates these recessions to the extent of the data.
- The user can opt to remove ongoing recessions from any plot. This is likeliest to be relevant in any long-term projection graphs, for which an ongoing recession continuing through, say, 2050 would not be reasonable.

The PR includes relevant documentation updates and adjusts the built-in recessions table to use the new format.


R/geom_recessions.R
```diff
@@ -435,41 +458,63 @@ GeomRecessionsText <- ggproto(
...
+        select(-index)
+
+      if (!quietly) {message("Successfully fetched from NBER")}
+
+      # Return recessions
+      recessions
+    },
+    error = function(cond){
+      if (!quietly) message("WARMING: Fetch or processing failed. `NULL` returned.")
+      return(NA)
```
matthewstern, line 516: If we're happy returning NA, the warning should say NA rather than NULL. However, I think NULL is probably the more accurate thing to return here. Also "warMing" :-)

R/geom_recessions.R
```diff
@@ -416,11 +439,11 @@ GeomRecessionsText <- ggproto(
 #' @return A tibble with the following variables: \itemize{ \item
 #'   \code{start_char, end_char}: Chr. Easily readable labels for the beginning
 #'   and end of the recession \item \code{start_num, end_num}: Double. Dates
-#'   expressed as years, with decimels referring to months. (e.g. April = 4/12 =
+#'   expressed as years, with decimals referring to months. (e.g. April = 4/12 =
 #'   .333) \item \code{start_date, end_date}: Date. Dates expressed in R
 #'   datetime format, using the first day of the specified month. }
 #'
-#' @source \url{https://www.nber.org/cycles/NBER chronology.xlsx}
+#' @source \url{https://www.nber.org/data/cycles 'cycle dates pasted.csv'}
```
matthewstern, line 446: we might want to clarify our definition of the `url` param now that we're using a different file than the standard one a user would find by just browsing the NBER website.

matthewstern (commented):
quick visual code scan looks great. will try to test/tinker in the coming days.

R/geom_recessions.R
dlcomeaux: Good catch on both.

R/geom_recessions.R
dlcomeaux: OK I've updated it. Although as an FYI, this is linked to on the business cycles page - it's the link to "JSON File Format" [here](https://www.nber.org/research/data/us-business-cycle-expansions-and-contractions).

dlcomeaux:
@matthewstern I made some additional tweaks to simplify the recessions table - moving the date calculation into the function, and improving the readability of the character field (into the "Mar 2021" format as opposed to a text string of "2021/03/01".

One change that might be worth it, but I don't know how to do, is shifting the filtering of ongoing recessions into `filter_recessions` - I just don't understand the mechanics of passing arguments into the ggproto object.

R/geom_recessions.R
```diff
@@ -196,41 +207,57 @@ geom_recessions <- function(xformat = "numeric",
...
 
   # unwrap recess_table from list
   recess_table <- recess_table[[1]]
 
-  # filter recessions correctly, based on xformat
+  # Add numeric version of dates
+  recess_table <- recess_table %>%
+    mutate(
+      start_num = lubridate::decimal_date(start_date),
+      end_num = lubridate::decimal_date(end_date))
```
matthewstern, line 254: Thinking more about this, if this does not happen in update_recessions like it used to, it should only happen if it needs to--eg if xscale is numeric. It would be easy to move this down a few lines into the if statement, and mutate rather than rename to get to start and end. 

Also if this is here we probably need to import and require lubridate? Not sure why check doesn't catch this.

matthewstern (approved):
Looks great to me! Would appreciate a bit of confirmatory testing of the small changes I made last night @dlcomeaux

DESCRIPTION
```diff
@@ -55,7 +55,6 @@ Suggests:
     knitr,
     lubridate,
     readxl,
```
nmpeterson, line 57: Can we remove readxl now that the NBER file being fetched is a CSV?

matthewstern: yes, sorry, I missed that one.

nmpeterson:
If we're targeting this for 1.2, but also want to include numerous other changes in that release, I think we should start a 1.2 dev branch to merge changes into, until they're all ready to go.

matthewstern:
> 
> 
> If we're targeting this for 1.2, but also want to include numerous other changes in that release, I think we should start a 1.2 dev branch to merge changes into, until they're all ready to go.

We're all over it! There isn't a draft PR yet for the branch but the branch exists:
image

DESCRIPTION
dlcomeaux: good catch

dlcomeaux:
All of these changes make sense to me, @matthewstern . Thanks for improving it!

Issue #26, opened 2020-02-14
improved x axis control in TS


matthewstern:
In time series, x labels should be integers. Also, in some situations we have published graphs where the first year is full ("2001") and those that follow are partial ('02, '03, '04, etc). We could make this an easy add (I've already found a source for the first half).

matthewstern:
Re establishing partial breaks, `scale_*_date()` and `scale_*_continuous()` in ggplot2 have arguments for `labels`, which according to their documentation can take a function that inputs the axes' breaks and returns a string of labels. It should be relatively easy to program a function that abbreviates certain dates. I would recommend that any function we program takes an argument that specifies which dates to not abbreviate. For example:

```
labels = abbreviate_dates(full_by_pos = c(1,8)) # don't abbreviate date 1 or 8
labels = abbreviate_dates(full_by_date = c(2000, 2010)) # don't abbreviate 2000 or 2010
```

dlcomeaux:
I found some code ([here](https://joshuacook.netlify.app/post/integer-values-ggplot-axis/)) that accomplishes the first and I'll take a crack at the second.

matthewstern:
@dlcomeaux  note that [this half-baked function](https://cmap-repos.github.io/cmapplot/reference/integer_breaks.html) has been a part of cmapplot for forever. It can be built upon or dumped if better tools are available.

dlcomeaux:
@matthewstern yes I found and am using that one, it seems to work fine.

PR #126, opened 2021-03-24
Allow for abbreviating years to two-digit versions

dlcomeaux:
This PR adds two related functions that allow users to abbreviate years on the x and y axes to their two-digit representations (e.g., 2008 to '08), while also allowing them to specify specific years or positions to keep at their full value, such as the first one or ones that start a new century. It addresses the issue raised in #26 .

The code in this is, I will admit, a bit strange, particularly on the date side. ggplot's methods of storing and handling breaks and labels are a bit opaque. I believe that I have properly accounted for some of the unusual bits - for example, when breaks are modified through things like `scales::breaks_pretty` ggplot somewhat inexplicably adds an `NA` to the list of breaks, requiring positions to be incremented up by one - but I think this could benefit from some further testing around edge cases.


dlcomeaux:
Also - this function works most cleanly right now when you are using the default case, keeping only the first value (i.e., inputting `labels = abbr_years_cont` or `labels = abbr_years_date`. 

However, if you want to customize the function, the only way I've found for it to work right now is to call it in the form `labels = function(i) abbr_years_cont(i,...)`. If anyone has any ideas on how to implement this more simply, that would be great - I couldn't find a way that ggplot would accept, but it's possible that it can be done.

matthewstern:
Thanks Daniel! Yet another deep, dark corner of ggplot.

I've just started to explore this, and wanted to report one finding: I believe the sometimes presence of leading and/or tailing NAs in the list of breaks represent (or enable) data to appear before or after the first and last breaks, respectively. 

For example, in this plot, the limits are larger than the breaks:
```
> pbuilt <- ggplot_build(
+     ggplot(cmapplot::grp_over_time, aes(x = year, y = realgrp, color = cluster)) +
+     geom_line() +
+     scale_x_continuous(breaks = scales::breaks_pretty(3)))

> pbuilt[["layout"]][["panel_params"]][[1]][["x"]][["limits"]]
[1] 2007 2017

> pbuilt[["layout"]][["panel_params"]][[1]][["x"]][["breaks"]]
[1]   NA 2010 2015   NA
```

But in this plot, I've set the first and last breaks to be the same as the limits:
```
> pbuilt <- ggplot_build(
+     ggplot(cmapplot::grp_over_time, aes(x = year, y = realgrp, color = cluster)) +
+     geom_line() +
+     scale_x_continuous(breaks = c(2007, 2010, 2014, 2017)))

> pbuilt[["layout"]][["panel_params"]][[1]][["x"]][["limits"]]
[1] 2007 2017

> pbuilt[["layout"]][["panel_params"]][[1]][["x"]][["breaks"]]
[1] 2007 2010 2014 2017
```

However, I'm not yet clear on how the custom labels function interacts with this as it gets factored in well before the "build" of the ggplot.

matthewstern:
Next observation, to your question about how this gets specified and needing to write `labels = function(i) abbr_years_cont(i,...)`: It looks to me like the trick might be that `labels` is able to take a function or a vector. But `f(i = 2010)` isn't a function, it's whatever the result of `f(i = 2010)` is. `f`, on the other hand, is a function.

For example, this plot removes 1 from each break very simply:

```
# function to subtract 1 from each break
fn <- function(breaks){breaks-1}

# breaks at `c(2007, 2010, 2014, 2017)`, but labeled as `c(2006, 2009, 2013, 2016)`
ggplot(cmapplot::grp_over_time, aes(x = year, y = realgrp, color = cluster)) +
    geom_line() +
    scale_x_continuous(breaks = c(2007, 2010, 2014, 2017), labels = fn)
```

The trick, if we want to specify some other variable, is returning a function that returns a function: 
```
# This `fn` returns a *function* that takes `breaks` as it's single input and subtracts n from it
> fn <- function(n){
+     return(function(breaks){breaks - n})
+ }

# breaks at `c(2007, 2010, 2014, 2017)`, but labeled as `c(2002, 2005, 2009, 2012)`
> ggplot(cmapplot::grp_over_time, aes(x = year, y = realgrp, color = cluster)) +
+     geom_line() +
+     scale_x_continuous(breaks = c(2007, 2010, 2014, 2017), labels = fn(5))
```

To get this to work the way we want to, and to avoid the NA handling issue you ran into, we might find it easiest to build a function that returns a function. Obviously, my simple subtraction experiment would need to evolve to into your abbreviation code. There may be best practices out there on the internet about how to do that? I haven't looked yet.

dlcomeaux:
OK @matthewstern this should address the syntax issues we discussed.

R/axis_handling.R
```diff
@@ -0,0 +1,151 @@
...
+    # how the breaks are handled internally by ggplot - they live in the following
+    # location of a ggplot chart 'a' that has been passed to ggplot_build()
+    #          ggplot_build(a)$layout$panel_params[[1]]$x$breaks
+    # Note that breaks for the y axis can be accessed similarly.
+    if (!is.null(full_by_year))
+      {full_by_year <- as.numeric(as.Date(paste0(full_by_year,"-01-01")))}
+
+    # Stop if breaks cannot be coerced to a date
+    tryCatch({as.Date(breaks)},
+             warning = function(w){stop("Breaks cannot be coerced to a date. Cannot abbreviate.")})
```
matthewstern, line 119: Looks like a limited amount of code is custom to date axes... is that right? Is it worth compressing this into a single function that avoids duplication feasible? it could take `dateaxis = FALSE` as an arg, for example.

R/axis_handling.R
```diff
@@ -0,0 +1,151 @@
...
+    lengths <- stringr::str_length(as.integer(breaks[which(!is.na(breaks))]))
+
+    # # Stop if the breaks are not in a four-digit format.
+    # if (!(min(lengths == 4) & max(lengths == 4))) {
+    #   stop("Breaks are not in a four-digit format. Cannot abbreviate.")
+    # }
+
+    # # Stop if breaks cannot be coerced to a number
+    # tryCatch({as.integer(breaks)},
+    #          warning = function(w){stop("Breaks cannot be coerced to a year. Cannot abbreviate.")})
```
matthewstern, line 77: Why is the above code commented out? It seems like, if it is unnecessary, the lengths calculation in line 68 is also unnecessary?

R/axis_handling.R
```diff
@@ -0,0 +1,151 @@
...
+      {full_by_year <- as.numeric(as.Date(paste0(full_by_year,"-01-01")))}
+
+    # Stop if breaks cannot be coerced to a date
+    tryCatch({as.Date(breaks)},
+             warning = function(w){stop("Breaks cannot be coerced to a date. Cannot abbreviate.")})
+
+    # Abbreviate all values - ggplot appears to correctly use the names of the
+    # breaks (which are the year values) when performing this operation on the
+    # breaks.
+    abbr <- paste0("'",substr(breaks,3,4))
```
matthewstern, line 124: it's weird that this works with `breaks` but below you need `names(breaks)`. I wonder, if you combine the two functions into one, whether it could be easiest and safer to do something at the top like:

```
# not sure these semantics are right:
if(dateaxis){breaks <- year(as.date(breaks))}
```

and then let the rest of the function work identically?

matthewstern (commented):
Have started to test. Getting an error in one testing case with the date version, which I have talked about offline with @dlcomeaux. A few comments for the time being, will look more later.

R/axis_handling.R
dlcomeaux: Yes, I will consolidate. It was separate because before turning it into a function factory, I wanted there to be a default version that would run for both.

R/axis_handling.R
dlcomeaux: Yeah I will play around with that. Some of the internal mechanics are opaque though, so it might stay as is.

R/axis_handling.R
dlcomeaux: Oh, it's commented because I had this in an earlier iteration and it was causing it to break. I'll have to adjust all of these anyway when consolidating the two.

R/axis_handling.R
dlcomeaux: I tried this, and was not able to get it to work. I could have been doing something wrong, but couldn't figure it out. Decided to stop as this is a rather weird rabbit hole and it seems to be working fine as is right now.

matthewstern (approved):
@dlcomeaux I got to do some extensive testing this afternoon and found a few ways I was able to simplify some of this underlying code:

- all breaks are now analyzed the same way. If `dateaxis == TRUE`, then breaks are converted to years. 
- new `stop` functionality: if `dateaxis == TRUE`, each break must be on the same calendar day, otherwise the labeling function would work but inaccurately describe the resulting graph
- the calculation of `actual_dates` via removing NAs has been removed. `na.omit()` and `na.rm = TRUE` have been used instead.

At this point, my testing indicates that this function works both inside `scale_x_continuous()`/`scale_x_date()` and outside of it, ie `abbr_years()(c(2010:2020))`.

Because I made some changes, I'd appreciate you taking another peek. But, if you think this is OK, let's merge it in!

dlcomeaux:
@matthewstern I think all your changes look good. Made one small tweak to comments. Thanks for improving it!

Issue #127, opened 2021-03-29
v1.2 font handling and output goals


matthewstern:
This issue is meant to combine the related issues #91, #116, and #120, which I will close after posting this. There may be more useful details in each of those discussions, but I've done my best to capture the most useful links and findings below.

**The situation:**
- our use of `svg()` does not export text (it converts text to shapes) and has not proven readable by Illustrator
- recent developments in packages `systemfonts`, `ragg`, and `svglite` make cross-platform, consistent graphics rendering with custom fonts possible
- our pkgdown website currently does not display Whitney fonts -- [I believe this is due to](https://github.com/CMAP-REPOS/cmapplot/issues/116#issuecomment-789058859) implementation of `ragg:agg_png()` (and therefore `systemfonts`) within the `pkgdown` package.
- due to the way we previously built fonts, [we changed](https://github.com/CMAP-REPOS/cmapplot/issues/117#issuecomment-809493874) `cmapplot_globals` to a named list rather than an environment, although storing these sorts of variables in an environment seems to be the [better practice](https://stackoverflow.com/a/12605694). We may be able to return the environment structure with the new fonts construction (which may enable an easy fix to #117). 

**Definite projects for v1.2:** (followed by possibly useful resources)
- [x] Implement the registering of Whitney fonts via systemfonts package and update font references throughout package to work accordingly. If this works well, also remove the sysfonts-based registry process. (**this will likely be a cooperative effort**)
  - https://cran.r-project.org/web/packages/systemfonts/systemfonts.pdf
  - See Matt's experiment [here](https://github.com/CMAP-REPOS/cmapplot/issues/120#issue-826484406)
- [x] Update `finalize_plot()` to use ragg-based raster functions (@sarahcmap) -> in progress in branch v1.2finalize
  - https://www.tidyverse.org/blog/2021/02/modern-text-features/#advanced-font-feature-support
  - https://ragg.r-lib.org/
- [ ] Update `finalize_plot()` to use `svglite()` instead of `svg()` (@sarahcmap) -> in progress in branch v1.2finalize
  - https://www.tidyverse.org/blog/2021/02/svglite-2-0-0/ 
  - https://svglite.r-lib.org/articles/fonts.html
- [x] Figure out a sane/reasonable way to notify users that they need to use `agg` as their RStudioGD backend, and/or manually set this for them (@nmpeterson)
  - https://ragg.r-lib.org/#use-ragg-in-rstudio
[...]

Issue #120, opened 2021-03-09
Update to systemfonts and ragg?


matthewstern:
We should discuss whether it makes sense to upgrade font handling to use the new package systemfonts and new raster devices from the ragg package.

Resources:

- https://www.tidyverse.org/blog/2021/02/modern-text-features/#advanced-font-feature-support
- https://cran.r-project.org/web/packages/systemfonts/systemfonts.pdf
- https://ragg.r-lib.org/

While systemfonts should be able to handle fonts natively to some degree, it looks like a good way to guarantee font handling is to check for the font by filtering the df returned by `system_fonts()`, and if found use `register_font()`. We could do this with all three fonts we use, e.g.:

```
library(systemfonts)
register_font("Whitney Medium", "C:\\Windows\\Fonts\\Whitney-Medium-Adv.otf")
registry_fonts()

>                                         path index         family       style weight italic features
> 1 C:\\Windows\\Fonts\\Whitney-Medium-Adv.otf     0 Whitney Medium     Regular normal  FALSE     NULL
> 2 C:\\Windows\\Fonts\\Whitney-Medium-Adv.otf     0 Whitney Medium        Bold   bold  FALSE     NULL
> 3 C:\\Windows\\Fonts\\Whitney-Medium-Adv.otf     0 Whitney Medium      Italic normal   TRUE     NULL
> 4 C:\\Windows\\Fonts\\Whitney-Medium-Adv.otf     0 Whitney Medium Bold Italic   bold   TRUE     NULL
```
... except use the path from the df returned by `system_fonts()`.

However, devices that don't communicate with systemfonts will not recognize these fonts. So, we'd need to address that across the board:

- upgrading to ragg devices for rasters solves this
- upgrading to svglite now solves this as of last month 
- PDFs ????
- RStudio plotting supports agg via [global settings override](https://ragg.r-lib.org/#use-ragg-in-rstudio) in v 1.4 and up.

I just upgraded RStudio (with IT help) and can confirm that after "registering" Whitney Medium as above...
```
ggplot() + 
    geom_text(
        aes(x = 0, y = 1, label = "the default is Arial"),
        size = 6
    ) + 
    geom_text(
        aes(x = 0, y = 0, label = "This is Whitney Medium"),
        family = "Whitney Medium",
        size = 6
    ) + 
    expand_limits(y = c(-1, 2)) + theme_void()
```
[...]

PR #128, opened 2021-04-05
Systemfonts and friends

matthewstern:
## Changes
- replaced sysfonts-based font lookup with systemfonts-based font "registration"
- did away with Calibri on PCs as fallback - just left it as default sans-serif font on all platforms
- updated vignettes to draw pngs with ragg devices
- converted `cmapplot_globals` from an exported list to an environment
- created three new functions: `get_cmapplot_globals`, `get_cmapplot_global`, and `set_cmapplot_global` to view and override entries within globals if needed. New documentation as needed.
- ~~moved the package startup message from `.onLoad` to `.onAttach` to avoid the note produced by `devtools::check` and conform to best practices.~~
- shifted `cmapplot_globals` creation into new file `cmapplot_globals.R`
- locally-built pkgdown site seems to work again!
- adjusted pkgdown action build to `gh-pages-test` branch on any PR, and to `gh-pages` on push to master.

Note that `finalize_plot` *saved outputs* won't work as this PR does not encapsulate the work in progress by @sarahcmap, but it should look right in the plots window if you have updated RStudio and [enabled an "agg" backend](https://www.tidyverse.org/blog/2021/02/modern-text-features/#using-ragg). 

## Testing requests:

- [x] ~~`check()` is producing an error at "rebuilding vignette outputs" about ragg. I believe I modified the vignettes correctly as per [here](https://www.tidyverse.org/blog/2021/02/modern-text-features/#using-ragg) and the output does seem to be working for me. Will others explore to see if they run into the same issue? (This may not be our problem?)~~
- [x] test fonts and vignette construction on Mac
- [x] test fonts, vignette builds, and global getting and setting via installation from github (vs load_all) (`devtools::install_github("CMAP-REPOS/cmapplot", "systemfonts", build_vignettes = TRUE)`)
- [x] test fonts on a machine where Whitney has been installed into a local user directory rather than `C:/Windows/fonts/`. I believe this will be the case if you installed Whitney without an admin password from IT.
*confirm here:*
image

## To dos this PR produces:

- [ ] I'm not sure the "new window" mode for finalize plot supports ragg. (As it is, this mode currently only works on a PC). We'll need to investigate whether there is a way to open a new window device with ragg, or remove the "new window" mode entirely from `finalize_plot`. @sarahcmap if this PR gets merged in would you be willing to explore this along with your other modifications to that function?
- [ ] Check to see if the use of systemfonts prevents breakage in the documentation examples for `theme_cmap()` and `finalize_plot()`. If these now do not cause errors, rewrite the example sections of these documentation pages to actually run examples.


sarahcmap (approved):
- On cmap laptop: 
  - Installed from GitHub using devtools.  
  - Vignettes and locally built pkgdown site looked good from a quick click-through, 
  - plot creation within RStudio with ragg backend looks good. 
- On personal mac: 
   - Installed with GitHub using devtools.  
   - Vignettes look good, did not try local pkgdown site
   - plotting in RStudio with ragg backend looks good (same as PC plot)
- I'll look into how/if the 'window' mode would work with ragg as part of finalize_plot.  I'll pull these changes into my v1.2finalize branch to make sure everything works together.

matthewstern:
> * On cmap laptop:
>   * plot creation within RStudio with ragg backend looks good.
> 

Thanks @sarahcmap, Mac confirmation with systemfonts is really exciting! Any chance the Whitney fonts on your cmap laptop are installed NOT in `c:\windows\fonts\`?

sarahcmap:
> Thanks @sarahcmap, Mac confirmation with systemfonts is really exciting! Any chance the Whitney fonts on your cmap laptop are installed NOT in `c:\windows\fonts\`?

They are in the usual folder (C:/Windows/Fonts), so can't comment on that item unfortunately!

nmpeterson:
I've taken a look on my Windows laptop; still have to review on Mac. Moved a few chunks of code around.

It looks like, if you don't explicitly change the default graphics rendering in RStudio to AGG, plots will produce errors like the following, which I... don't *love.* Does everybody else get these errors, too?

> Warning messages:
> 1: In grid.Call(C_textBounds, as.graphicsAnnot(x$label), x$x, x$y,  :
>   font family not found in Windows font database
> 2: In grid.Call.graphics(C_text, as.graphicsAnnot(x$label), x$x, x$y,  :
>   font family not found in Windows font database
> 3: In grid.Call.graphics(C_text, as.graphicsAnnot(x$label), x$x, x$y,  :
>   font family not found in Windows font database
> 4: In grid.Call(C_textBounds, as.graphicsAnnot(x$label), x$x, x$y,  :
>   font family not found in Windows font database
> 5: In grid.Call(C_textBounds, as.graphicsAnnot(x$label), x$x, x$y,  :
>   font family not found in Windows font database

If we go forward with this, we'll have to update the readme and the installation vignette to make it very clear that AGG has to be used.

nmpeterson:
Also, I just ran `pkgdown::build_site()` and everything is in Arial.

sarahcmap:
What font is the pkgdown site supposed to be again?  Mine looks like it's in Lato...

nmpeterson:
@sarahcmap the text content of the website doesn't matter, that's controlled by the theme set in _pkgdown.yaml. I'm referring specifically to the fonts in the R graphics that get built for the site, which *should* all be in Whitney. Easiest way to check, in my opinion, is to open the cookbook vignette and compare the fonts in the cmapplot graphics against the original Comms versions.

sarahcmap:
Ah ok.  Mine looks like Whitney based on the 'g' in Chicago.  

image

sarahcmap:
> It looks like, if you don't explicitly change the default graphics rendering in RStudio to AGG, plots will produce errors like the following, which I... don't _love._ Does everybody else get these errors, too?

Yes when i switch back to default

R/cmapplot_globals.R
```diff
@@ -1,27 +1,27 @@
...
+# Set up default font handling
+# (Note: overridden by .onLoad() if Whitney is available)
 cmapplot_globals$use_whitney <- FALSE
 cmapplot_globals$font <- list(
   strong = list(family = "sans", face = "bold"),
   regular = list(family = "sans", face = "plain"),
   light = list(family = "sans", face = "plain"))
 
-# establish font sizes
+# Set common font sizes (bigpts)
```
matthewstern, line 12: These are actually points (1/72.27 inch, I believe) - everything in the package except for fonts uses bigpts, so that fractions of inches are clean. fonts use real points, so that on import to illustrator they are read as 11 instead of 10.9, etc.

matthewstern (commented):
Thanks @nmpeterson. Love the changes.

> It looks like, if you don't explicitly change the default graphics rendering in RStudio to AGG, plots will produce errors like the following, which I... don't love. Does everybody else get these errors, too?

In my mind, cmapplot 1.2 and up would *require* ragg as to avoid these error messages (which I also get). I imagine you missed, it (it was a long message), but over in #127 I asked if you would take a look coming up with a decent way to notify users that ragg drawing in the plot window is required upon package load, or setting it for them. Either of them would be easy to do in `.onLoad()` or `.onAttach()` -- I demonstrated how the package could check or change the value [here](https://github.com/CMAP-REPOS/cmapplot/issues/120#issue-826484406) -- it's just a question of how to go about it. 

> Also, I just ran pkgdown::build_site() and everything is in Arial.

Interesting, when I run `build_site()` it works, as it did for @sarahcmap. No chance `pkgdown` or `ragg` is out of date?

nmpeterson:
> In my mind, cmapplot 1.2 and up would _require_ ragg as to avoid these error messages (which I also get). I imagine you missed, it (it was a long message), but over in #127 I asked if you would take a look coming up with a decent way to notify users that ragg drawing in the plot window is required upon package load, or setting it for them. Either of them would be easy to do in `.onLoad()` or `.onAttach()` -- I demonstrated how the package could check or change the value [here](https://github.com/CMAP-REPOS/cmapplot/issues/120#issue-826484406) -- it's just a question of how to go about it.

Yes, sorry, I did miss that. I'll work on adding that to this PR.

> Interesting, when I run `build_site()` it works, as it did for @sarahcmap. No chance `pkgdown` or `ragg` is out of date?

pkgdown 1.6.1, ragg 1.1.2. Pretty sure those are the latest.

dlcomeaux:
I tested this on my work computer, which has Whitney installed outside of the Fonts folder. It works fine for plotting and exporting (although not for window generation). However, the kerning issues on PNG exports are as bad (if not worse) than they were in the previous iteration: 
image

R/cmapplot_globals.R
nmpeterson: Thanks, I'll change to pts

nmpeterson:
@dlcomeaux when you run `library(cmapplot)`, is that after installing this specific branch (i.e. `devtools::install_github('cmap-repos/cmapplot', ref='systemfonts')`)? Otherwise, it's just loading the old implementation.

matthewstern:
> > Interesting, when I run `build_site()` it works, as it did for @sarahcmap. No chance `pkgdown` or `ragg` is out of date?
> 
> pkgdown 1.6.1, ragg 1.1.2. Pretty sure those are the latest.

Well, considering the pkgdown site gets built by a VM, I'm going to suggest we let it be for now and hope that the VM's build acts like mine and Sarah's machine and not yours, @nmpeterson? Or, is there an easy way to have the github action build an alternate copy of the pkgdown site based on the `version1.2` branch once this branch gets merged there?



> I tested this on my work computer, which has Whitney installed outside of the Fonts folder. It works fine for plotting and exporting

This is great news! Seems like `systemfonts` is doing it's job on Windows, Windows with user-specific Whitney fonts, and on @sarahcmap's mac. Hurrah!



> However, kerning is better when loaded through `library(cmapplot)`

@dlcomeaux This is weird but, I think, not super relevant then. considering that `devtools::load_all` is only supposed to be an approximation of package loading, if it works via `library()` then the issue seems moot to me.

nmpeterson:
Test pkgdown site appears to also use Arial

line-1

nmpeterson:
Confirmed: `systemfonts::system_fonts()` on the virtual machine is only finding fonts in /System/Library/Fonts (which we can't write to), and not the CMAP fonts that get installed in ~/Library/Fonts. Not sure what the path forward is...

dlcomeaux:
> 
> 
> @dlcomeaux when you run `library(cmapplot)`, is that after installing this specific branch (i.e. `devtools::install_github('cmap-repos/cmapplot', ref='systemfonts')`)? Otherwise, it's just loading the old implementation.

Yes, I am running the latest version of this branch.

dlcomeaux:
> @dlcomeaux If you are running `library()` on this branch's version of the package as Noel suggests, and it works with no kerning issues, than this seems like a non-issue to me.

Agreed, it's not a problem we need to solve. I am curious why there is that discrepancy, but not worth the rabbit hole I'm sure it would entail.

matthewstern:
Thanks for digging, @nmpeterson. This looks messy! [Looks to me](https://github.com/CMAP-REPOS/cmapplot/blob/gh-pages-test/articles/cookbook_files/figure-html/area-1.png) like we're still building with Arial? 

Your discovery that others have this issue is reassuring. Do you want to let this settle for now, or do you have more tricks up your sleeve?

nmpeterson:
I'm going to tinker with the pkgdown action a little more, but I have very little confidence that I will succeed. If not, maybe we should put this PR on ice and update the pkgdown action in master to build with an older version of pkgdown?

nmpeterson:
Ok, I'm stuck. I tried registering the fonts manually with `systemfonts::register_font()` (and modifying the `use_whitney` logic to search both `systemfonts::system_fonts()` and `systemfonts::registry_fonts()`). This means that when the pkgdown site is built, it is doing so in an environment where `use_whitney == TRUE`, and yet the graphics are still being generated with Arial.

matthewstern:
@nmpeterson any chance this could be an otf vs ttf issue? Looks like when you dig into Petr Bouchal's supposed success with `register_font` that he is registering ttfs:

[His rmd code:](https://github.com/petrbouchal/pagetest/blob/master/index.Rmd)
```
`{r index-7}
ptrr::register_plexsans()
ptrr::register_plexsans_condensed()
`

`{r index-8}
systemfonts::registry_fonts()
`
```

[His personal package:](https://github.com/petrbouchal/ptrr/blob/master/R/fonts-ragg.R):
```
egister_plexsans_condensed <- function(family = "IBM Plex Sans Condensed", features = NULL, warn = TRUE) {

  if(family %in% systemfonts::system_fonts()[["family"]]) {
    if(warn) usethis::ui_warn(" Family {usethis::ui_value(family)} already exists in system font registry. Skipping.")
    invisible(FALSE)
  } else {

    systemfonts::register_font(family,
                               plain = system.file("fonts", "plex",
                                                   "IBMPlexSansCondensed-Regular.ttf",
                                                   package = "ptrr"),
                               italic = system.file("fonts", "plex",
                                                    "IBMPlexSansCondensed-Italic.ttf",
                                                    package = "ptrr"),
                               bold = system.file("fonts", "plex",
                                                  "IBMPlexSansCondensed-Bold.ttf",
                                                  package = "ptrr"),
                               bolditalic = system.file("fonts", "plex",
                                                        "IBMPlexSansCondensed-BoldItalic.ttf",
                                                        package = "ptrr")
    )
    invisible(TRUE)
  }
```

PS this flags a future to do if we ever get this working... 
 

- [ ] we need to replicate his logic to not re-register Whitney if it has already been registered. The font registry seems to not expire with the R session, and re-registering may be creating duplicates on my system?

nmpeterson:
> @nmpeterson any chance this could be an otf vs ttf issue? Looks like when you dig into Petr Bouchal's supposed success with `register_font` that he is registering ttfs
> PS this flags a future to do if we ever get this working...

Seems unlikely. He is using TTFs, yes, but it is also not working for him. (If you look at his site, it's using a non-default font, but that font is _not_ IBM Plex Sans.

>     * [ ]  we need to replicate his logic to not re-register Whitney if it has already been registered. The font registry seems to not expire with the R session, and re-registering may be creating duplicates on my system?

I disagree with this. I think cmapplot itself should not be running `register_font()` (because we don't know where they'd be located), and instead just rely on `systemfonts` finding them itself. The exception would be the GitHub action, where we have to register them ourselves. While this would be a problem with using `register_font()` on our own computers, every single run of a GitHub action is a blank slate. There is zero persistence of any files or environment settings from one run to another. The VM is literally created and destroyed as part of the run :)

matthewstern:
> I disagree with this. I think cmapplot itself should not be running `register_font()` (because we don't know where they'd be located), and instead just rely on `systemfonts` finding them itself. The exception would be the GitHub action, where we have to register them ourselves. While this would be a problem with using `register_font()` on our own computers, every single run of a GitHub action is a blank slate. There is zero persistence of any files or environment settings from one run to another. The VM is literally created and destroyed as part of the run :)

OK, so I think I misread the details of the `systemfonts` package. I had originally thought that ragg might not necessarily find fonts if they are installed in atypical directories (e.g. not `c:/windows/fonts/`), which is why I programmed `.onLoad` in this branch to `register_fonts()` using paths extracted by `system_fonts()`. Upon rereading, I suspect that my logic is a catch-22: if `system_fonts()` finds it, there is no need to register it. If this is in fact the case, this isn't hard to change.

@nmpeterson I don't really understand the `pkgdown_test.yaml` file. Is it really "installing" the font or just copying it to a local directory on the VM? If the purpose of `register_fonts()` is to allow use of fonts in R that aren't actually "installed" to the system (and it appears to be, see test below), is it worth changing the `pkgdown_test.yaml` file to copy the otf/ttf files to a directory that might be less problematic, then registering them manually? 

## registering an uninstalled font and using it in ggplot
```
library(systemfonts)
library(ggplot2)

register_font("foobar", "C:\\Users\\mstern\\Desktop\\CMAPSerifBold.ttf")

ggplot() + 
  geom_text(
    aes(x = 0, y = 1, label = "the default is Arial"),
    size = 6
  ) + 
  geom_text(
    aes(x = 0, y = 0, label = "This is foobar (CMAP Serif)"),
    family = "foobar",
    size = 6
  ) + 
  expand_limits(y = c(-1, 2)) + theme_void()
```
image

nmpeterson:
> @nmpeterson I don't really understand the `pkgdown_test.yaml` file. Is it really "installing" the font or just copying it to a local directory on the VM?

On a typical macOS setup, this is a distinction without a difference. Copying them to one of a few different specific locations is the means by which they are effectively installed. But the GHA macOS setup doesn't seem to recognize fonts installed in the user-specific fonts folder.

> If the purpose of `register_fonts()` is to allow use of fonts in R that aren't actually "installed" to the system (and it appears to be, see test below), is it worth changing the `pkgdown_test.yaml` file to copy the otf/ttf files to a directory that might be less problematic, then registering them manually?

I have updated pkgdown_test.yaml to register directly from the cloned cmap-fonts repo, rather than copying to ~/Library/Fonts first. We'll see what happens.

matthewstern:
@nmpeterson this is ready for your re-review. I think [this view](https://github.com/CMAP-REPOS/cmapplot/pull/128/files/b721deec5e8ffafdfdaa59e938367245d6b687cb..90803ab7a73ba1744eff635c233e1fc73fc1fcc5) of changed files is probably the one to go with. It goes from commit b721dee, which was when you started to explore solutions to the pkgdown issue, to present. 

Some notes: 
- font registration in all circumstances seems to be a must. Because r packages including grid and ggplot2 specify fonts via their "family", fonts like "Whitney-Book" are unavailable to systemfonts because they are not the primary font in the Whitney family, nor are they specifiable via a font-face lookup. The systemfonts::register_variant() function reveals some about this. 
- font registration is not sticky - it must be re-performed on every R session
- the solution here was that font registration must be handled within `.onLoad()` for vignettes to use the right fonts, because vignettes are run in their own, clean environment. 
- `onLoad()` needed to be correctly specified to register the necessary whitney variants in both Windows and OSX contexts. In Windows, the filepath is discovered using `systemfonts::system_fonts()`, which finds the fonts even though they are unavailable due to the issue described above. However, in OSX environments (or at least, in the GHA VM), that method does not work, so the ~\Library\Fonts directory is now searched for fontfiles that contain Whitney in their filename.

If this looks and tests out OK to you, in my mind these are the remaining to dos:
- [ ] [some sort of push](https://github.com/CMAP-REPOS/cmapplot/pull/128#pullrequestreview-628066319) to use agg backend in `.onLoad()` or `.onAttach()`. 
- [ ] remove the test pkgdown action
- [ ] Would it be helpful to keep the pkgdown test branch around? I found [the ability to render html pages directly from github ](https://htmlpreview.github.io/) to be quite useful in finding a solution to this problem. Is it easy to modify the primary pkgdown action to push to a branch dynamically? (e.g. gh-pages for a commit to master, gh-pages-test for a PR whose destination is master?)

R/cmapplot.R
```diff
@@ -22,23 +22,53 @@
...
+  # If font registry contains whitney core, set use_whitney == TRUE
+  if(!get("use_whitney", envir = cmapplot_globals) & .Platform$OS.type != "windows"){
+
+    user_dir <- paste0(Sys.getenv("HOME"), "/Library/Fonts")
+
+    if(dir.exists(user_dir)){
+      whitney_fonts <- list.files(user_dir, full.names = TRUE) %>%
+        as.data.frame() %>%
+        rlang::set_names("path") %>%
+        dplyr::filter(stringr::str_detect(path, "Whitney")) %>% # Will error for username of "Whitney"
```
nmpeterson, line 57: To avoid the potential error for unluckily named users, can't the `str_detect()` check for "Whitney-" instead of "Whitney"?

nmpeterson:
@matthewstern so nice to see it (almost) working!

One new issue we need to resolve now is getting the italic fonts correctly registered as a pair with the non-italicized variant of the same weight by using the italic argument in systemfonts::register_font(). Right now, if you try to italicize text in a plot (e.g. in the "title and captioning" section of the finalize vignette), it uses the un-italicized font. It's not even applying software-based slanting. For the same reason, there is no way to apply bold text.

I'm pretty sure the Whitney specification in `systemfonts::system_fonts()` (regardless of OS -- just not on the GitHub Action VMs) already handles this correctly for users who have the fonts installed, so I think if possible we should only use `register_fonts()` if the required fonts are not already present in `system_fonts()`.

 --------

Regarding the pkgdown test branch, I'm not sure whether it's possible to configure a single action to change its output branch based on context, but we can definitely have two separate actions, with the standard one building to `gh-pages` on commit to `master` and the test version building to `gh-pages-test` on pull request to master.

matthewstern:
Thanks for the review @nmpeterson!

> I'm pretty sure the Whitney specification in `systemfonts::system_fonts()` (regardless of OS -- just not on the GitHub Action VMs) already handles this correctly for users who have the fonts installed, so I think if possible we should only use `register_fonts()` if the required fonts are not already present in `system_fonts()`.

Can you demonstrate with any sample code getting R to access "Whitney-Book" without registering the font? I have not been able to.  I made the determination that this is likely due to this level of detail is stored in `systemfonts::system_fonts()`'s `name` column, as opposed to `family`. (notice that when you `register_font()`, the name you specify is listed as as `family`.) It is my belief that this is why `systemfonts::register_variant()` exists. But, I definitely stand ready to be corrected and would happily recode to avoid registration where possible.

> One new issue we need to resolve now is getting the italic fonts correctly registered as a pair with the non-italicized variant of the same weight by using the italic argument in systemfonts::register_font(). Right now, if you try to italicize text in a plot (e.g. in the "title and captioning" section of the finalize vignette), it uses the un-italicized font. It's not even applying software-based slanting. For the same reason, there is no way to apply bold text.

This is a good catch! I hadn't tested for this. Did you check if **bold** works? Presuming the problem here is italics only, I believe this solution is conceptually easy but one of those things I'm not sure I'm good enough to code efficiently. I believe we'd want to `reshape_wide` the `whitney_fonts` df (generated either from `system_fonts` or `list.files`) based on whether "Italic" is present in the name to create two paths (one regular, one italic). Then pass this to `register_font` walking over 3 variables rather than 2. If that sounds like the right tactic, I'm happy to give it a go.

> Regarding the pkgdown test branch, I'm not sure whether it's possible to configure a single action to change its output branch based on context, but we can definitely have two separate actions, with the standard one building to `gh-pages` on commit to `master` and the test version building to `gh-pages-test` on pull request to master.

OK, I'm cool with this.  I was wondering if [Github environment variables](https://docs.github.com/en/actions/reference/environment-variables) could do this, but if it's not already obvious to you it hardly seems worth experimenting with.

matthewstern:
A follow-up on accessing fonts within the Whitney family without `register_font()`. This example seems to demonstrate that, at least on my machine, "Whitney-Bold" is the default font used when the "Whitney" family is specified, and that font lookup by `name` is unavailable without font registration.

```
library(systemfonts)
library(ggplot2)

register_font("TestFont", "C:\\Windows\\Fonts\\Whitney-Bold-Adv.otf")

ggplot() + 
  geom_text(
    aes(x = 0, y = 1, label = "This is `Whitney` when specified by family"),
    family = "Whitney",
    size = 6
  ) + 
  geom_text(
    aes(x = 0, y = 0, label = "This is Whitney Bold, registered as 'testfont'"),
    family = "TestFont",
    size = 6
  ) + 
  geom_text(
    aes(x = 0, y = -1, label = paste(
      "This is an attempt at accessing Whitney-Bold by name.",
      "'Whitney-Bold' is not a family, so this defaults to sans-serif",
      sep = "\n"
      )),
    family = "Whitney-Bold",
    size = 6
  ) + 
  expand_limits(y = c(-2, 2)) + theme_void()
```

image

matthewstern:
This is ready for re-review! 

This branch now... 
- registers the three needed fonts with all necessary specified variants (plain, bold, italic, bolditalic). The fonts are named overtly in `.onLoad`, and if any of the fonts are not found, `use_whitney` will not be set to `TRUE`. At least on my PC, it searches for systemwide fonts ("C/Windows/Fonts") and user-specific fonts. 
- checks for RStudio version and graphics backend setting, with necessary actions and warnings
- updates the pkgdown Github action to dynamically deploy to `gh-pages` on a push to `master`, or `gh-pages-test` on any update to an open PR. 

FWIW, the fonts I've chosen to stand in as the various needed variants. There is some discretion with what is used for bold, I picked what made the most sense to me:
```
      # register preferred strong font (Whitney Semibold), with variants
      systemfonts::register_font(
        name = cmapplot_globals$preferred_font$strong,
        plain = find_path("Whitney-Semibold-Adv", whitney_paths),
        bold = find_path("Whitney-Black-Adv", whitney_paths),
        italic = find_path("Whitney-SemiboldItal-Adv", whitney_paths),
        bolditalic = find_path("Whitney-BlackItal-Adv", whitney_paths)
      )

      # register preferred regular font (Whitney Medium), with variants
      systemfonts::register_font(
        name = cmapplot_globals$preferred_font$regular,
        plain = find_path("Whitney-Medium-Adv", whitney_paths),
        bold = find_path("Whitney-Bold-Adv", whitney_paths),
        italic = find_path("Whitney-MediumItal-Adv", whitney_paths),
        bolditalic = find_path("Whitney-BoldItal-Adv", whitney_paths)
      )

      # register preferred light font (Whitney Book), with variants
      systemfonts::register_font(
        name = cmapplot_globals$preferred_font$light,
        plain = find_path("Whitney-Book-Adv", whitney_paths),
        bold = find_path("Whitney-Semibold-Adv", whitney_paths),
        italic = find_path("Whitney-BookItal-Adv", whitney_paths),
        bolditalic = find_path("Whitney-SemiboldItal-Adv", whitney_paths)
      )
```

nmpeterson (approved):

nmpeterson:
Works as advertised on my CMAP laptop

dlcomeaux:
I tried installing the latest version of this branch onto my CMAP laptop and was not able to get it to work. To confirm what I was supposed to do:
- I have updated RStudio to the latest version
- I changed graphics handling to AGG
- I installed the `systemfonts` branch using `devtools::install_github()`
- I loaded the `cmapplot` library.

When I do so, I got an error during installation, specifically relating to semibold:
```
*** arch - i386
Error : Font 'Whitney-Semibold-Adv' not found.
*** arch - x64
Error : Font 'Whitney-Semibold-Adv' not found.
** testing if installed package can be loaded from final location
*** arch - i386
Error : Font 'Whitney-Semibold-Adv' not found.
*** arch - x64
Error : Font 'Whitney-Semibold-Adv' not found.
** testing if installed package keeps a record of temporary installation path
* DONE (cmapplot)
```
This is a bit confusing to me, because when I run the `systemfonts` code in `cmapplot.R`, the system is able to find the path for semibold just fine. And, I don't think that semibold is installed any differently than the other two faces. Any ideas on what could be causing the issue? Perhaps there's an issue with the font, rather than with the R code (although the font seems to work just fine within Word)? Open to ideas, I'm at a loss.

dlcomeaux (approved):

matthewstern:
Thanks for the thorough testing and tweak @dlcomeaux.

This shouldn't cause any new errors I can think of because the way the filename is calculated: ` result <- grep(paste0("(\\\\|/)", filename, ".[ot]tf$"), paths, value = TRUE)` is quite conservative, as it goes from the last `/` to the end of the string, so the only possible matches are multiple installations of the same file or both an otf and ttf version of the same font. In either of these situations, returning an arbitrary filepath seems like an acceptable solution.

BTW I believe the repeat errors you experienced on installation were because you had `build_vignettes = TRUE`. Most of our vignettes begin with `library(cmapplot)`, which would then reproduce the error. It would have been interesting to confirm, while the error was occurring, that the vignettes had built, but had rendered in Arial.

matthewstern:
Speaking of which, @sarahcmap @nmpeterson @nmpeterson did any of you install .ttfs? I've only tested with .otfs and am curious to confirm that everything renders similarly with .ttfs.

matthewstern:
Speaking of which, @sarahcmap @nmpeterson @nmpeterson did any of you install .ttfs? I've only tested with .otfs and am curious to confirm that everything renders similarly with .ttfs.

dlcomeaux:
> Thanks for the thorough testing and tweak @dlcomeaux.
> 
> 
> 
> This shouldn't cause any new errors I can think of because the way the filename is calculated: ` result <- grep(paste0("(\\\\|/)", filename, ".[ot]tf$"), paths, value = TRUE)` is quite conservative, as it goes from the last `/` or `\` to the end of the string, so the only possible matches I can think of are multiple installations of the same file or both an otf and ttf version of the same font. In either of these situations, returning an arbitrary filepath seems like an acceptable solution.
> 
> 
> 
> BTW I believe the repeat errors you experienced on installation were because you had `build_vignettes = TRUE`. Most of our vignettes begin with `library(cmapplot)`, which would then reproduce the error. It would have been interesting to confirm, while the error was occurring, that the vignettes had built, but had rendered in Arial.

Is `build_vignettes=T`the default? I didn't set it explicitly. And to your other question, I have .OTFs

nmpeterson:
@matthewstern I have not tried TTFs. I.T. only gave me OTFs. I created the TTF files using a website that does the format conversion because at one point when I was originally working on the font specification I thought OTFs may not work with a package I was considering using. However, I have no idea how reliable that conversion process was -- the resulting TTFs could conceivably have issues due to their dubious origin, so probably best to stick with the OTFs. (Please feel free to test, though, either with the Whitney TTFs or another typeface for which you can obtain "official" TTFs.)

matthewstern:
Thanks Noel. @dlcomeaux just tested with the TTFs you shared with us some time ago and indicated they worked. I had forgotten that you had manufactured them, though! Good to know, and I will stick to OTFs going forward. TBH I'm just glad I'm finally figuring out how to speak in regex a bit.

PR #131, opened 2021-05-11
Allow for discrete versions of gradients

dlcomeaux:
This is a relatively simple PR - it just adds the gradients to the list of allowed color palettes for the discrete scales. I also updated the documentation and vignettes accordingly. I could be missing something, so please let me know @gritzenthaler if there's some reason why this wouldn't work in practice.

One thing I was thinking about as additional work might be to find a way to "recenter" a diverging discrete palette around some non-mid-point using factors (I can see instances where this might come up in a mapping context, i.e., you have five bins above zero and two below - the current code would center it around the 4th factor, no matter the value). However, I couldn't see an easy way to do that, since the discrete scales don't seem to have a rescaling option easily available.

Open to other thoughts - I thought at the very least this adds the requested functionality and doesn't preclude a more sophisticated approach later on.

### DIFF
--- a/R/colors_continuous.R
+++ b/R/colors_continuous.R
@@ -141,19 +141,36 @@ mid_rescaler2 <- function(mid) {
 #' @export
 cmap_fill_continuous <- function(palette = "seq_reds",
                                  reverse = FALSE,
+                                 discrete = FALSE,
                                  middle = 0,
                                  ...) {
     if (substr(palette,1,3) == "div") {
-        ggplot2::scale_fill_gradientn(
-            colours = cmap_pal_continuous(palette, reverse = reverse)(256),
-            rescaler = mid_rescaler2(middle),
-            ...
-        )
+        if (discrete) {
+            ggplot2::discrete_scale(
+                "fill", "cmap_palettes",
+                palette = cmap_pal_discrete(palette, reverse = reverse),
+                ...
+            )
+        } else {
+            ggplot2::scale_fill_gradientn(
+                colours = cmap_pal_continuous(palette, reverse = reverse)(256),
+                rescaler = mid_rescaler2(middle),
+                ...
+                )
+        }
     } else {
-        ggplot2::scale_fill_gradientn(
-            colours = cmap_pal_continuous(palette, reverse = reverse)(256),
-            ...
-        )
+        if (discrete) {
+            ggplot2::discrete_scale(
+                "fill", "cmap_palettes",
+                palette = cmap_pal_discrete(palette, reverse = reverse),
+                ...
+            )
+        } else {
+            ggplot2::scale_fill_gradientn(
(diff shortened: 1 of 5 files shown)


dlcomeaux:
Note - this passed check on my machine. It seems there is an issue with the pkgdown deployment to the test branch, I would suggest we re-run later in case it's a temporary glitch (unless @nmpeterson or @matthewstern have encountered this before / know what it might be a result of?). Here's the error:

```
  x Failed to parse Rd in abbr_years.Rd
ℹ there is no package called ‘textshaping’ 

Execution halted
Error: Process completed with exit code 1.
```

matthewstern:
FWIW, rerunning the previously-successful pkgdown action on the most recent commit to the version1.2 branch (ebdbb66) creates a similar error situation and fails. Comparing the GHA log for a successful vs unsuccessful run shows entirely different response to the "install dependencies" code. I am hoping these failures have something to with a package update elsewhere, and this will resolve itself automatically in a few days.

nmpeterson:
@dlcomeaux FYI, you can re-run a failed check without a new commit. Just go to the Actions tab, open the failed action you want to re-run, then click "Re-run jobs"

nmpeterson:
Also, regarding the current error with the textshaping package, just keep an eye on its [CRAN page](https://cran.r-project.org/web/packages/textshaping/index.html). Once the macOS binary version (currently 0.3.3) matches the version listed at the top (0.3.4), the action should work. The way the action is configured, it tries to install the latest version of every dependency, but if the binaries haven't been updated then they can't be installed.

nmpeterson:
Also, regarding the current error with the textshaping package, just keep an eye on its [CRAN page](https://cran.r-project.org/web/packages/textshaping/index.html). Once the macOS binary version (currently 0.3.3) matches the version listed at the top (0.3.4), the action should work. The way the action is configured, it tries to install the latest version of every dependency, but if the binaries haven't been updated then they can't be installed.

nmpeterson:
Also, regarding the current error with the textshaping package, just keep an eye on its [CRAN page](https://cran.r-project.org/web/packages/textshaping/index.html). Once the macOS binary version (currently 0.3.3) matches the version listed at the top (0.3.4), the action should work. The way the action is configured, it tries to install the latest version of every dependency, but if the binaries haven't been updated then they can't be installed.

dlcomeaux:
@nmpeterson the binaries have been updated, and everything seems to be in order again.

matthewstern (approved):
I am OK with this changes but would like @gritzenthaler to confirm before merging.

gritzenthaler:
I've been rather inactive on this project lately but getting to this now! I'll review it by the end of the day.

gritzenthaler (approved):
It looks great, thanks Daniel!

PR #132, opened 2021-05-13
Vignette image fix

dlcomeaux:
Preliminary PR to fix the vignette margin image problem. I'm not sure if this currently works - it builds it locally, but the old code did as well. @nmpeterson - I think you had previously shared it was possible to do a test build of the site, right (but couldn't find where)? How do we do that again?

### DIFF
--- a/pkgdown/_pkgdown.yml
+++ b/pkgdown/_pkgdown.yml
@@ -55,3 +55,7 @@ articles:
   - finalize
   - colors
   - cookbook
+
+resource_files:
+- man/figures/margins1.png
+- man/figures/margins2.png
--- a/vignettes/finalize.Rmd
+++ b/vignettes/finalize.Rmd
@@ -199,8 +199,8 @@ There is a fairly long list of possible margins that can be customized using the
 
 Here, the margins are visualized as they impact a finalized plot that has a sidebar:
 
-<img src="./../man/figures/margins1.png" width="100%" style="display: block; margin: auto;">
+<img src="./../reference/figures/margins1.png" width="100%" style="display: block; margin: auto;">
 
 Here, the same margins are visualized as they impact a finalized plot with no sidebar:
 
-<img src="./../man/figures/margins2.png" width="100%" style="display: block; margin: auto;">
+<img src="./../reference/figures/margins2.png" width="100%" style="display: block; margin: auto;">
--- a/vignettes/finalize.Rmd
+++ b/vignettes/finalize.Rmd
@@ -199,8 +199,8 @@ There is a fairly long list of possible margins that can be customized using the
 
 Here, the margins are visualized as they impact a finalized plot that has a sidebar:
 
-<img src="./../reference/figures/margins1.png" width="100%" style="display: block; margin: auto;">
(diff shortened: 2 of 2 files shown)


nmpeterson:
Every time a pull request is created (or updated), the pkgdown action will attempt to build the website into the [**gh-pages-test** branch](https://github.com/CMAP-REPOS/cmapplot/tree/gh-pages-test)

dlcomeaux:
Ah ok @nmpeterson - when I clone the current version, which appears to be based on the commit, the HTML files work when I open them in my browser. However, that also appears to be the case for previous code (I downloaded an earlier set of code from the test branch). So, I'm not sure whether this has had any impact, or how to test it except pushing it into the main branch. This is definitely not a part of the repo I particularly understand, though, so, open to other ideas.

nmpeterson:
I finally had some time to look at this in some detail. Everything looks good to me. Vignettes seem to load the images correctly in both the pkgdown build and the R install (i.e. `pkgdown::install(build_vignettes=TRUE)` => `library(cmapplot)` => `browseVignettes("cmapplot")`).

nmpeterson:
I'm fine with it. Not sure if @matthewstern wants to review first.

matthewstern:
I'm ok with trying this although it's worth noting that previewing here does not work:
https://raw.githubusercontent.com/CMAP-REPOS/cmapplot/gh-pages-test/man/figures/margins1.PNG

nmpeterson:
@matthewstern GitHub seems to have case-sensitive URLs. This works: https://raw.githubusercontent.com/CMAP-REPOS/cmapplot/gh-pages-test/man/figures/margins1.png

matthewstern:
@nmpeterson you have, I believe, identified the actual underlying problem. The [currently live pkgdown website](https://cmap-repos.github.io/cmapplot/articles/finalize.html#many-many-margins-1) links to `https://cmap-repos.github.io/cmapplot/man/figures/margins1.PNG` but the file that exists is `https://cmap-repos.github.io/cmapplot/man/figures/margins1.png`. This is despite the fact that the vignette itself links to the lowercase png: ``
      
This suggests to me that the fix here just merged is actually not going to work. I would guess this error came to pass originally because of a pkgdown update that is for some reason capitalizing the file extension. Identifying the pngs as resources as @dlcomeaux has done here might be part of the solution, but I don't think it's going to take us all the way.

matthewstern:
@dlcomeaux I'm wondering the simplest solution here may actually be to rename the files in the github directory to `.PNG`, and edit the reference in the vignette so that it also looks for a capitalized extension. Theoretically, the external files code you added would no longer be necessary ([pkgdown suggests](https://pkgdown.r-lib.org/reference/build_articles.html#external-files) that automatic detection should work, and `resource_files` is just a backup). 

Would you be willing to give this a try?

nmpeterson:
@dlcomeaux before you try @matthewstern's suggestion, let's review my new PR (#133). The repo already had the files as uppercase. I renamed them to lower case using command line git, so I believe no further changes should be necessary.

dlcomeaux:
Thanks all - just getting to this, seems like everything is in order. Thanks!

Issue #129, opened 2021-04-08
Margins image is not displaying in vignettes


dlcomeaux:
For some reason, at least on my computer, the margins pngs are not displaying in the finalize vignette. Is anyone else having this issue?
![image](https://user-images.githubusercontent.com/58892984/114091333-2e98d000-987e-11eb-8500-2f6a3ef924f0.png)

nmpeterson:
Most recent [pkgdown run](https://github.com/CMAP-REPOS/cmapplot/runs/2015464881?check_suite_focus=true)'s log says (under "Deploy package"):

> Copying 'man/figures/margins1.PNG' to 'reference/figures/margins1.PNG'
> Copying 'man/figures/margins2.PNG' to 'reference/figures/margins2.PNG'

Meanwhile, the vignette includes the following HTML to display the images:

```
Here, the margins are visualized as they impact a finalized plot that has a sidebar:

<img src="./../man/figures/margins1.png" width="100%" style="display: block; margin: auto;"/>

Here, the same margins are visualized as they impact a finalized plot with no sidebar:

<img src="./../man/figures/margins2.png" width="100%" style="display: block; margin: auto;"/>
```

I guess those file references should just be changed to say `./../reference` instead of `./../man`? Would this create any issues in terms of viewing vignettes locally?

nmpeterson:
[This seems relevant.](https://pkgdown.r-lib.org/reference/build_articles.html#external-files)

nmpeterson:
Fixed by #133.

PR #133, opened 2021-05-28
Change case of PNG file extensions

nmpeterson:
On my PC, the file extensions were showing as uppercase. I used `git mv` on the command line to rename them as lowercase. Hopefully this resolves the issue (#129).


nmpeterson:
After nuking the contents of `gh-pages-test` and re-running the pkgdown action, both sets of images (the originals in man/figures and the copied versions in reference/figures) are now displayed in GitHub with lowercase extensions. We may also have to nuke `gh-pages` when version 1.2 gets merged into master.

matthewstern:
Looks to me like this addressed the issue:

The previewed render of the html page here...
https://htmlpreview.github.io/?https://github.com/CMAP-REPOS/cmapplot/blob/gh-pages-test/articles/finalize.html

...shows the image and looks for the file as a lowercase extension, which is available here...
https://raw.githubusercontent.com/CMAP-REPOS/cmapplot/gh-pages-test/man/figures/margins1.png


would you guys agree that we should remove the resource files linking merged via #132, as that seems likely to be superfluous?

nmpeterson:
I think it's fine to leave it, just in case pkgdown's automatic detection stops working in the future for any reason. From [the docs](https://pkgdown.r-lib.org/reference/build_articles.html#external-files):

> pkgdown differs from base R in its handling of external files. When building vignettes, R assumes that vignettes are self-contained (a reasonable assumption when most vignettes were PDFs) and only copies files explicitly listed in .install_extras. pkgdown takes a different approach based on rmarkdown::find_external_resources(), and it will also copy any images that you link to. **_If for some reason the automatic detection doesn't work, you will need to add a resource_files field to the yaml metadata_**

matthewstern (approved):

matthewstern:
> After nuking the contents of `gh-pages-test` and re-running the pkgdown action,

@nmpeterson did the rename not take effect until you nuked the branch?

nmpeterson:
@matthewstern correct. `pkgdown::deploy_to_branch(clean = TRUE)` does not seem to work exactly as advertised.

matthewstern:
> 
> 
> @matthewstern correct. `pkgdown::deploy_to_branch(clean = TRUE)` does not seem to work exactly as advertised.
> 
> **Edit:** to be clear, the rename of the files in man/figures was successful, but the copies in reference/figures remained uppercase until I nuked them.

It doesn't particularly matter, but it's worth noting that I don't believe the `reference/figures/margin#.png` file actually gets used. While the pkgdown site does look for things like our logo in that folder (`https://cmap-repos.github.io/cmapplot/reference/figures/logo.png`) the way we've constructed the html image tag for margins does not seem to get overwritten:

html tag in Rmd file:
``

link in pkgdown HTML file:
`https://cmap-repos.github.io/cmapplot/man/figures/margins1.PNG`

which is not not:
`https://cmap-repos.github.io/cmapplot/reference/figures/margins1.PNG`

it may be that there is a different/less complex way to specify the `img src` that would be interpreted differently by the pkgdown process.

Issue #119, opened 2021-03-02
unable to use cmapplot continuous color palettes for fill aesthetic


tkoyejo:
I wanted to apply the built-in CMAP continuous palette to produce a continuous gradient fill aesthetic in a stacked bar chart. When I apply `cmap_fill_continuous(palette = "seq_blues")` to a stacked bar chart, I get the following error: "Error: Discrete value supplied to continuous scale". If I switch to `cmap_color_continuous(palette = "seq_red_purple")`, the code is ignored and the default ggplot colors are used.

I ended up having to use the ggplot color brewer, which worked. The example output I seek is attached. 
`scale_fill_brewer(palette = "Blues")` 

![figure17_title](https://user-images.githubusercontent.com/22479581/109708527-2227a600-7b61-11eb-8bde-bbdc23be7660.png)

matthewstern:
@gritzenthaler, this is a bit of a different color application than we had originally scoped and tested. Is this code modifiable to enable the application of a continuous palette to a discrete set of categories?

gritzenthaler:
Yes I believe we talked about this sort of application a while ago, particularly in regards to choropleth maps. I'm sure it's possible, so I'll look into it once I have some time in the coming week.

matthewstern:
Is this the same issue as #70?

tkoyejo:
If I'm reading this right, sounds like the same issue to me.

matthewstern:
@gritzenthaler for the time being I've assigned you to both issues and added both to the v1.2 milestone. Feel free to close one or the other if you determine them to be duplicates.

PR #135, opened 2021-06-21
Color tweaks

matthewstern:
- `cmap_palettes` and `cmap_gradients` no longer exist as exported lists. Now, there is a `palettes` tibble in `cmapplot_globals` that is constructed upon package load that contains all palettes.
- `viz_gradient`, `viz_palette`, and various palette applier functions have been adjusted to pull from this new tibble
- a new function `fetch_pal` has been added to utilities.R to make it easier to get details about a palette, such as its colors vector.
- the colors vignette has been updated, including shifting the color lists from the viz function help page to the colors vignette.
- `cmap_*_discrete()` functions issue a warning if you use the race palette with them, as users should really be using that palette only with `cmap_*_race()`.


matthewstern:
For review. Although these were fairly small changes, they ended up spanning a wide variety of files. 

For documentation changes, see especially:
- [new `viz_*` help page](https://htmlpreview.github.io/?https://github.com/CMAP-REPOS/cmapplot/blob/gh-pages-test/reference/viz_palette.html) (combination of old `viz_gradient` and `viz_palette`)
- [Updated colors article/vignette](https://htmlpreview.github.io/?https://github.com/CMAP-REPOS/cmapplot/blob/gh-pages-test/articles/colors.html), with all palettes visible in a new lower section.

R/utilities.R
```diff
@@ -140,3 +138,45 @@ safe_grobHeight <- function(grob, unitTo = "bigpts", valueOnly = TRUE){
...
+#' # Confirm that "reds" is a sequential palette
+#' fetch_pal("reds", which = "sequential", return = "exists")
+#'
+#' @export
+fetch_pal <- function(pal,
+                      which = c("discrete", "sequential", "divergent"), #unique(cmapplot_globals$palettes$type),
+                      return = c("colors", "type", "exists")){
+  # basics
+  name <- type <- NULL
+  return <- match.arg(return)
```
matthewstern, line 165: could add, if we wanted: `which <- match.arg(which, unique(cmapplot_globals$palettes$type), several.ok = TRUE)`

nmpeterson: Would this display the allowed options in the help files?

matthewstern: This is separate, but a response to your tweak. I'm suggesting we leave your change to the function defaults, but add a `match.arg` to confirm what the user calls is acceptable.

dlcomeaux (approved):
Made a few minor tweaks and added something to the vignette about using sequential palettes with discrete scales. Looks good to me overall. Nice work @matthewstern.

vignettes/colors.Rmd
```diff
@@ -148,8 +148,15 @@ nms <- dplyr::filter(get_cmapplot_global("palettes"),  type == "sequential")$nam
 purrr::walk(nms, viz_gradient)
 ```
 
+If you use a sequential palette for a discrete scale, the package will automatically choose colors from across the selected gradient and interpolate additional ones if needed. For example, the `blues` palette is shown below, interpolating nine colors from the initial palette list of seven.
+
+```{r show_sequential_discrete, echo = FALSE, fig.asp = 300/670}
+nms <- fetch_pal("blues")
+viz_palette("blues",num = 9)
```
matthewstern, line 155: @dlcomeaux I'm OK with this, I think. I appreciate the value of this clarification, but I don't *love* that it's in the section that is intended as a menu of options as opposed to a description of how something works. What would you think about moving this up to the `### Discrete scales` section up top, and combining the `viz_palette` call that demonstrates the interpolation with an actual example of a ggplot with a discrete scale of a continuous palette? Something like [Timi's here](https://github.com/CMAP-REPOS/cmapplot/issues/119#issue-820350735). 

Also, does the `nms` construction here have a purpose?

dlcomeaux: Oops, the nms was a holdover. That should be cut.

nmpeterson (approved):
Looks good to me

gritzenthaler (approved):
This looks good! Putting all the palettes in globals will make it much easier to update them later.

Issue #142, opened 2022-02-11
Remove unused race/ethnicity categories from legend


nmpeterson:
Currently, if you create a chart using `cmap_color_race()`, the resulting legend will display all five categories (white, black, hispanic, asian, other), even if one of those categories is not present in the data/chart:

![race_legend](https://user-images.githubusercontent.com/4383272/153654732-d230b640-6712-4d27-8956-8216bb59e89b.png)

I think there are two possible approaches for addressing this:

1. Remove any default labels for the race/ethnicity categories, and only construct the palette from the label names that are explicitly supplied to `cmap_color_race()`. So `cmap_color_race(white = "White", black = "Black")` would yield a 2-item legend instead of 5 (with "hispanic", "asian" and "other" appearing by default). If none of the parameters were specified, an error would be returned saying that at least one must be.
2. Add a new `remove_missing` boolean parameter. If it defaulted to `FALSE`, then `cmap_color_race(white = "White", black = "Black")` would still produce the same 5-item legend as the current function, but `cmap_color_race(white = "White", black = "Black", remove_missing = TRUE)` would then produce a 2-item legend.

I don't have strong feelings one way or the other. I'd be particularly interested to hear @gritzenthaler's thoughts on this one.

PR #143, opened 2022-07-19
remove unused race categories from legend

leotorres114:
Decided to try to implement Noel's first approach from [Issue 142](https://github.com/CMAP-REPOS/cmapplot/issues/142), which seems a bit more intuitive to the end user, e.g. provide which race categories are available the data and `cmap_fill_race()` / `cmap_color_race()` should automatically exclude those not provided. 

`match.call()` was used because `cmap_fill/color_race` is passing all of its arguments to `make_race_palette`, which uses that list of arguments to build the palette. if `match.call()` wasn't used, the arguments passed to `make_race_palette` would be unevaluated `symbol` class (better explained [here](https://zhuchcn.github.io/projects/posts/en/2019-03-22/#matchcall-a-pitall-in-r)), which wouldn't allow `make_race_palette` to build a palette from the arguments passed to it. Most resources pointed to the need to use `match.call()` for this scenario, but maybe there's a better way to address this?

`do.call()` is just executing the `make_race_palette` with a list of the arguments (names and values) provided to `cmap_fill/color_race()`. 

To build the palette, `make_race_palette` loops through the names of `race_palette` - if that name is included in the names of the argument list `passed`, it adds the argument list value of `passed` as the index and hex value of `race_palette` as the value to an empty character vector, `pal`.


R/colors_race.R
```diff
@@ -8,21 +8,31 @@
 #'
 #' @noRd
 make_race_palette <- function(white, black, hispanic, asian, other) {
+    if (missing(white) & missing(black) & missing(hispanic) & missing(asian) & missing(other)) {
+      stop("You must specify at least one race category", call. = FALSE)
```
dlcomeaux, line 12: So, I think that this actually needs to be handled a bit differently - the function was originally developed so that, in the absence of any arguments, it would default to looking for "white" and "asian" and "hispanic" and so on. I believe how you've constructed it here requires people to put it in, even if the values are appropriately named for the default. Could you update the code so that it maintains the default behavior?

R/colors_race.R
```diff
@@ -8,21 +8,31 @@
...
+    }
+
+    passed <- unlist(as.list(match.call())[-1]) # vector of args actually passed
+
     race_palette <- fetch_pal("race")
-    if (!missing(white)) { names(race_palette)[1] <- white }
```
dlcomeaux, line 12: I think this section is going to need to be maintained in some way, although I'm not sure on the sequencing. But the goal would be that, in cases where people do have default values, the function remains functional, even without specification.

R/colors_race.R
```diff
@@ -41,15 +51,17 @@ make_race_palette <- function(white, black, hispanic, asian, other) {
 #' @describeIn cmap_fill_race For fill aesthetic
 #' @export
 cmap_fill_race <- function(white, black, hispanic, asian, other) {
-    race_palette <- make_race_palette(white, black, hispanic, asian, other)
+    .args <- as.list(match.call()[-1])
+    race_palette <- do.call(make_race_palette, .args)
```
dlcomeaux, line 55: This works well, nicely done.

dlcomeaux (changes requested):

dlcomeaux (changes requested):
When the code is done, please also update documentation (using the RStudio tools to do that, Ctrl + Shift + D should do the trick) and make any necessary updates to the vignettes. Thanks!

dlcomeaux (changes requested):
And, we'll want to update the version number and note the changes in News.md

R/colors_race.R
```diff
@@ -8,21 +8,31 @@
...
+
+    return (pal)
+}
 
 #' Apply official CMAP race/ethnicity chart colors to ggplot2 aesthetics
 #'
 #' Pick the function depending on the aesthetic of your ggplot object (fill or color).
 #' Specify your dataset's unique race factor names (as case-sensitive strings) in the arguments.
-#' All categories are optional in case your dataset does not have some of them.
+#' All categories are optional in case your dataset does not have some of them, however, at least one
```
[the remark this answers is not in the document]
leotorres114: Hmm, not sure we can do this unless we pass the dataframe itself as a function parameter (to be able to determine if any parameters are `%in%` the race column). ggplot's `scale_fill_manual` does [not](https://github.com/tidyverse/ggplot2/blob/7ddb6d91513d3bfff43e2cee916eb819d4772a3e/R/scale-manual.r#L150) check if there are valid values in the dataset, it just maps those that are valid -- unless no parameters were specified, in which case it errors.

R/colors_race.R
leotorres114: In the case where the dataframe already has the default values, the code has been updated for the function to return the default race palette.

R/colors_race.R
leotorres114: Makes sense, code has been updated to maintain the default behavior!

leotorres114:
@dlcomeaux documentation, vignettes, and News.md have been updated accordingly. Let me know if you have any thoughts about my reply to [your](https://github.com/CMAP-REPOS/cmapplot/pull/143#discussion_r930401988) comment above.

dlcomeaux:
> @dlcomeaux documentation, vignettes, and News.md have been updated accordingly. Let me know if you have any thoughts about my reply to [your](https://github.com/CMAP-REPOS/cmapplot/pull/143#discussion_r930401988) comment above.

Sorry for the very long delay. This solution works, and I agree, seems more trouble than it's worth to go after a more complex solution.

dlcomeaux:
@leotorres114  this works, I will update documentation and merge.

dlcomeaux:
@EthanJantz @amcadamsCMAP question for you - this is failing in the pkgdown part of the run, with an error that I don't recognize. Are either of you familiar with what might cause this?

leotorres114:
@dlcomeaux it looks like the gh actions workflow file is outdated. it can't find `r-lib/actions@master` because it's been updated ([reference](https://github.com/r-lib/actions#releases-and-tags)). I created another PR to update the workflow file, #145 , with the correct version. updating that first should (hopefully) take care of that error.

PR #150, opened 2023-02-20
Palette interpolation

### LINKED ISSUE #149: Prevent palette interpolation for some cases
Generating a discrete color palette using `cmap_*_discrete` will interpolate colors from outside of a defined palette in cases where the number of colors used is less than the number of colors in the full palette. I think this is due to the use of `grDevices::colorRampPalette` as the palette function in `colors_discrete.R`. This interpolation is useful in cases where the number of colors needed is larger than the size of the palette or for continuous scales, but for cases where the number of colors needed is less we should expect that the palette function will simply pull from the palette in order, and not interpolate using `grDevices::colorRampPalette`. Below are the results of calling `viz_palette` on the "friday" palette for different values of `num`. 

image

I'm unsure if this is intended behavior or not. @nmpeterson do you have any insight?

dlcomeaux:
Initial code ideas to fix #149 . I haven't fully tested this out yet, but wanted to share what I thought might work. From some quick testing, I think it does the trick - although we would want to talk about what the ideal behavior would be for something like this. @nmpeterson @EthanJantz  @MMenninger flagging for your input. As it stands, this would:

- Use a discrete palette in order of colors if the palette needed up to the total number of colors
- If the palette needs more colors, it would interpolate, but would flag that this goes outside of CMAP design standards
- For gradients, it would use the old behavior (interpolating so that the end and the beginning of the gradient remain the same)

### DIFF
--- a/R/colors_discrete.R
+++ b/R/colors_discrete.R
@@ -87,11 +87,16 @@ cmap_pal_discrete <- function(palette = "prosperity", reverse = FALSE, gradient 
     return(
       function(n) {
 
+        # For gradients, interpolate to make a continuous gradient with
+        # consistent end points
         if (gradient == TRUE) {
           grDevices::colorRampPalette(pal)(n)
         }
 
+        # For everything else (discrete palettes)...
         else {
+          # If the number of colors exceeds the palette, interpolate but issue a
+          # warning
           if (n > length(pal)) {
 
             message(paste("WARNING: The palette you selected has",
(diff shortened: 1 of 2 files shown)


dlcomeaux:
On reflection I think it would probably make more sense to have the gradient behavior be keyed to a list of gradient palettes. I'll try to do that when I have a moment, but curious if others have thoughts before building that in.

EthanJantz:
I'm not a fan of having a gradient parameter in a discrete palette function. Users shouldn't be expecting to get a gradient from a discrete palette function, and I think that this would get confusing when trying to select the correct color/fill scale function from the package. If we want to have a single master palette for the package that we can use for both discrete and continuous data maybe we should consider dropping the distinction in our package and providing a parameter for a general color/fill scale function. Just my two cents.

dlcomeaux:
This is quite old and we will probably want to just handle as part of the larger update of color palettes (for discussion with Comms). Will leave this in draft for now, but flagging that we may shift to the approach that Ethan identified in the linked issue.

PR #151, opened 2023-03-03
change checkout and cache to v3

EthanJantz:
The pkgdown action runs on pull request or a push to `master`. This pull request should trigger it and let us know if this addresses issue 146.


dlcomeaux (approved):
Looks good to me - I found a couple other instances where the workflows were using an older version.

PR #157, opened 2023-10-30
Geom pandemics

jmaltman:
created geom_pandemics.R (not ready)
updated geom_recessions.R with new URL
updated sysdata.rda
created pandemics.rda
created R script to create pandemics.rda


jmaltman:
When trying to run geom_pandemics, this error occurs: 
Error in `geom_pandemics()`:
! Problem while setting up geom.
ℹ Error occurred in the 1st layer.
Caused by error in `UseMethod()`:
! no applicable method for 'mutate' applied to an object of class "character"
This seems to be an issue from line 219 of the geom_pandemics code. Not sure why this is happening since the same code (minus name changes) exists in geom_recessions and that function works

dlcomeaux:
@jmaltman heads up, I had a bit of time and was able to get the function to work. I did **not** address the `geom_recessions` data update link. Could you play around with that using the new format, ideally using the stable geoJSON link that they have now migrated to rather than the csv/xlsx approach we used before?

dlcomeaux:
It would also be better practice for the code if we streamlined/reduced duplication between geom_recessions and geom_pandemics. This can certainly be done, and ideally we would have a consolidated documentation too - but woudl take some work and require some changes to arguments (the most important one of which would be changing `update_recessions` and `update_pandemics` to a consistent `update` parameter).

dlcomeaux:
OK @jmaltman I would appreciate if you could see whether this works on your system, including in edge cases. I think it's a functional PR at this stage. When we are ready, we will need to increment the cmapplot version number and add a change log.

jmaltman:
Testing this function out with the update piece of the function to insert my own pandemics data, I found two errors come up.

1) The labels for the made-up pandemics did not change to reflect the name of these imaginary pandemics. Instead, each of them kept the label "COVID-19 Pandemic" in the chart.
2) For an imaginary pandemic that was ongoing (where ongoing = TRUE in the data), if no end date was added to the ongoing pandemic, then it would not populate in the chart. If I gave it an imaginary date in the future, then it appeared, but if I left the end date as NA, then it did not appear when trying to plot it.

dlcomeaux:
> 

Hmm, I am not able to replicate the bug for the labels. When I add a new pandemic table by copying the existing one and adding a new row, it shows up fine for me with the accurate label. Could you paste the code you used to yield this bug?

The ongoing one *requires* an end date by the way that the function currently works. You could add some code to do that automatically, if you wanted, but in the recessions table a user supplied one wouldn't work unless you put a far out date on it.

jmaltman:
@dlcomeaux, I created my own data frame instead of adding to the existing pandemic table.

new_df <- data.frame(start_char = c("Jan 1981", "Apr 2009", "Jun 2015"),
                     end_char = c("Jan 1989", "Aug 2010", NA),
                     start_date = c("1/20/1981", "4/25/2009", "6/15/2015"),
                     end_date = c("1/20/1989", "9/10/2010", NA),
                     ongoing = c(FALSE, FALSE, TRUE),
                     name = c("Reagan", "Swine Flu", "Trump")
                     )

new_df <- 
  new_df %>% 
  mutate(start_date = as.Date(start_date, format = "%m/%d/%Y"),
         end_date = as.Date(end_date, format = "%m/%d/%Y"))

ggplot(df, mapping = aes(x = year_dec, y = value)) +
  geom_pandemics(update = new_df) +
  geom_line(aes(color = var)) +
  scale_x_continuous("Year") +
  theme_minimal()

dlcomeaux:
@jmaltman OK this has identified a few things for me.
- One issue with this is the `NA` code - can you try what happens on your system if you give that one an end date? Do the labels work properly then?
- When I run that, it gives me an error that it may not be properly formatted, although it still outputs the geoms
- I think I know what is happening with the labels. The label is being extracted as a part of the geom before the update function is run, which means that it pulls from the built in labels.

